PR #2, opened 2020-08-06
chore: set up app config

### LINKED ISSUE #1: set up and config eslint, commitlint, prettier, and husky

hagoptaminian:
add eslint, husky, prettier, commitlint, .npmrc, package.json, .gitignore

closes #1


.eslintrc
```diff
@@ -0,0 +1,58 @@
+{
```
juancarlosfarah, line 1: Can you check if this matches the one in Graasp Dssktop?

package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "graasp-insights",
+  "version": "1.0.0",
+  "main": "index.js",
+  "repository": "https://github.com/graasp/graasp-insights.git",
```
juancarlosfarah, line 5: Use ssh+https. There's an example in one of our repos.

package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "graasp-insights",
+  "version": "1.0.0",
```
juancarlosfarah, line 3: Let's start with version 0.1.0.

package.json
```diff
@@ -0,0 +1,29 @@
...
+  "main": "index.js",
+  "repository": "https://github.com/graasp/graasp-insights.git",
+  "author": "React EPFL",
```
juancarlosfarah, line 6: Let's put Graasp here.

package.json
hagoptaminian: I checked graasp-api, -service-analytics, -app-moodle-analytics, -app-data-visualization, -app-timer, -research, and -desktop; several of them don't have a "repository" key, and the only one that's different is the graasp-desktop one. I used the one in there.

.eslintrc
hagoptaminian: it doesn't. specifically: desktop only uses `eslint-config-airbnb`, `eslint-config-prettier`, and `eslint-plugin-mocha`, while graasp-research is missing the latter but has the first 2 *plus* `eslint-plugin-import`, `eslint-plugin-jsx-a11y`, `eslint-plugin-react` (and corresponding amends to `.eslintrc`). since this is a react app (even if on desktop) I assumed we should follow the pattern of graasp-research. but i can amend it to be exactly like graasp-desktop. @juancarlosfarah let me know!

juancarlosfarah: Indeed it should have the React part too. Let's see what @pyphilia says.

pyphilia: I think it should work like this, you may add `eslint-plugin-mocha` as well since you included it in `env` already.

hagoptaminian: done!

PR #4, opened 2020-08-10
3/electron react

pyphilia:
This is a huge PR adding `electron` and `react` dependencies and necessary files to start the app.


.gitignore
```diff
@@ -3,3 +3,30 @@
...
+
+# vim settings
+.rootdir
+
+# exclude logs
+*.log
+
+# production
+build/*
+!build/favicon.ico
```
hagoptaminian, line 32: empty line at end of file missing

public/app/config/config.js
```diff
@@ -0,0 +1,29 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
hagoptaminian, line 1: interesting - this is flagged because electron is a dev dependency. wonder if it means we should adjust our eslint settings

pyphilia: I think it is better to have the warning (and disable it for electron) in case of other misused libraries.

public/electron.js
```diff
@@ -0,0 +1,226 @@
...
+const macFileSubmenu = [{ role: 'close' }];
+const standardFileSubmenu = [
+  {
+    label: 'About',
+    click: () => {
+      openAboutWindow({
+        // asset for icon is in the public/assets folder
+        base_path: escapeEscapeCharacter(app.getAppPath()),
+        icon_path: path.join(__dirname, ICON_PATH),
+        copyright: 'Copyright © 2019 React',
```
juancarlosfarah, line 105: Can we try and make the name dynamic? Also, let's use © 2020 EPFL & Graasp Association. If we can do it in two lines:

```
Copyright © 2020 EPFL
Copyright © 2020 Graasp Association
```

public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "Graasp",
```
juancarlosfarah, line 2: Maybe just `Insights` or `Graasp Insights` too?

src/App.js
```diff
@@ -0,0 +1,59 @@
+import React, { Component } from 'react';
+import { HashRouter as Router, Route, Switch } from 'react-router-dom';
+import { withStyles } from '@material-ui/core';
+import { withTranslation } from 'react-i18next';
+import { MuiThemeProvider } from '@material-ui/core/styles';
+import { HOME_PATH } from './config/paths';
+import theme from './theme';
+import Home from './components/Home';
+
+const styles = () => ({
```
juancarlosfarah, line 10: I don't think we're using the `styles` here.

juancarlosfarah: And if we will later, we should use camelCase for the keys.

src/components/common/Header.js
```diff
@@ -0,0 +1,75 @@
...
+  appBarShift: {
+    width: `calc(100% - ${DRAWER_WIDTH}px)`,
+    marginLeft: DRAWER_WIDTH,
+    transition: theme.transitions.create(['margin', 'width'], {
+      easing: theme.transitions.easing.easeOut,
+      duration: theme.transitions.duration.enteringScreen,
+    }),
+  },
+  menuButton: {
+    marginLeft: 12,
```
juancarlosfarah, line 28: Can we comment these fixed numbers? Or use constants?

src/components/common/Loader.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactLoading from 'react-loading';
+
+const Loader = ({ type }) => (
+  <div className="Loader">
+    <ReactLoading type={type} color="#5348d3" />
```
juancarlosfarah, line 7: Factor out colour? Or use `theme`.

package.json
```diff
@@ -13,22 +14,76 @@
     "Hagop Taminian"
   ],
   "license": "AGPL-3.0",
+  "dependencies": {
```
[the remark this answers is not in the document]
pyphilia: Yes I installed all the last dependencies, even for `electron`, `spectron`, etc

PR #7, opened 2020-08-26
test: set up spectron

pyphilia:
This PR uses `"spectron": "11.0.0"` because the version `11.1.0` contains breaking changes and throws an error for every test.

### DIFF
--- a/src/config/selectors.js
+++ b/src/config/selectors.js
@@ -1,3 +1,3 @@
 export const HOME_MENU_ITEM_ID = 'homeMenuItem';
 export const DRAWER_BUTTON_ID = 'drawerButton';
-export const QUIT_MENU_ITEM_ID = 'closeMenuItem';
+export const QUIT_MENU_ITEM_ID = 'quitMenuItem';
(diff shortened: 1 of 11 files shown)


src/config/selectors.js
```diff
@@ -1,2 +1,3 @@
 export const HOME_MENU_ITEM_ID = 'homeMenuItem';
 export const DRAWER_BUTTON_ID = 'drawerButton';
+export const QUIT_MENU_ITEM_ID = 'closeMenuItem';
```
juancarlosfarah, line 3: Maybe consistency between `quit` and `close`?

test/menu.test.js
```diff
@@ -0,0 +1,32 @@
...
+      app = await createApplication();
+    }),
+  );
+
+  afterEach(function () {
+    return closeApplication(app);
+  });
+
+  it(
+    'Displays correct navigation',
```
juancarlosfarah, line 24: Try to keep `it` blocks starting with lower case, to differentiate from `describe` blocks. Not important, just a tip.

juancarlosfarah (changes requested):
Thanks for adding this. Just some minor recommendations. Otherwise all good from my end. Check with @hagoptaminian for merging.

hagoptaminian (approved):
this looks good to me! ran `yarn mocha` locally and the tests are working

PR #12, opened 2020-09-09
feat: set up load file component

hagoptaminian:
*incomplete* code committed for draft PR purposes


hagoptaminian:
@pyphilia I committed this incomplete code to get your feedback, specifically on the failure to get the node `fs` module to work. I generally followed the code you have in graasp-desktop, but without the exact clean folder/file/separation-of-concerns that you have. My main goal was to get comfortable with electron and its two processes. In the end, I was able to send a message from the LoadFile component to the electron.js main process and `console.log` the contents of that message. **However** I was _not_ able to get the node `fs` module to work to create a copy of the file whose path was sent; I get an error along the lines of `fs.createCopy is not a function`. If you know what I might be missing, please let me know.

src/utils/browse.js
```diff
@@ -0,0 +1,21 @@
+import {
+  SHOW_LOAD_SPACE_PROMPT_CHANNEL,
+  RESPOND_LOAD_SPACE_PROMPT_CHANNEL,
+} from '../config/channels';
+
+// eslint-disable-next-line import/prefer-default-export
+export const showBrowsePrompt = (handleFileLocation) => {
```
pyphilia, line 7: I think it is a too specific function (load only datasets) to be put in `utils`. In `graasp-desktop` we instead go for more specific listeners (a show prompt for each purpose) instead of a generic one triggered by specific "actions". So I would rather move `options` in the listener directly, and have `showBrowsePrompt` in the `LoadFile` component. Let me know what you think.

pyphilia (changes requested):
It looks good, you're almost there!

public/electron.js
```diff
@@ -231,3 +239,13 @@ app.on('activate', () => {
 ipcMain.on('load-page', (event, arg) => {
   mainWindow.loadURL(arg);
 });
+
+ipcMain.on(CREATE_FILE_COPY, (event, args) => {
+  console.log(`event is ${event} and args is ${args.fileLocation}`);
+  fs.createCopy(args.fileLocation, 'newFile.json', (err) => {
```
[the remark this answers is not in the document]
hagoptaminian: you're right, it worked :) don't know where I got `fs.createCopy` 😶

public/electron.js
```diff
@@ -34,7 +40,7 @@ const createWindow = () => {
     show: false,
     movable: true,
     webPreferences: {
-      nodeIntegration: false,
+      nodeIntegration: true,
```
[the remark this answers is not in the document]
hagoptaminian: I thought this was what was preventing the `fs` module from running correctly, but it wasn't. restored it to `false`

src/components/LoadFile.js
```diff
@@ -0,0 +1,75 @@
...
+          value={fileLocation}
+          type="text"
+        />
+        <Button
+          //   id={LOAD_SUBMIT_BUTTON_ID}
+          variant="contained"
+          onClick={handleCopy}
+          color="primary"
+          //   className={classes.button}
+          //   disabled={!fileLocation.endsWith('.zip')}
```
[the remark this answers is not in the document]
hagoptaminian: of course, in `src/utils/browse.js` I set the format to `.json`, but I will uncomment this line and change it to `.json`

src/utils/browse.js
hagoptaminian: Makes sense. I moved this code into the `handleBrowse` function in the component, and removed the intermediate `showBrowsePrompt` function. However, not sure what you mean by moving the `options` to the listener (?). I left it in `handleBrowse`.

public/app/config/channels.js
```diff
@@ -1,3 +1,9 @@
 // this file needs to use module.exports as it is used both by react and
 // electron make sure this file is identical in both src/config/channels.js
 // and public/app/config/channels.js
+
+module.exports = {
+  SHOW_LOAD_SPACE_PROMPT_CHANNEL: 'prompt:space:load:show',
+  RESPOND_LOAD_SPACE_PROMPT_CHANNEL: 'prompt:space:load:respond',
+  CREATE_FILE_COPY: 'file:copy',
```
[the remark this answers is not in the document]
hagoptaminian: Agree, would be nice to capture the 'import' aspect of this action. I went with `CREATE_FILE_COPY: 'file:import'`, but let me know if there is a preferred convention (you would know better)

src/components/LoadFile.js
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  return (
+    <Main fullScreen>
+      <FormControl>
+        <Typography variant="h4" color="inherit" style={{ margin: '2rem' }}>
+          {t('Load a File')}
+        </Typography>
+        <Button
+          // id={LOAD_BROWSE_BUTTON_ID}
```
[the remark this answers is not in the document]
hagoptaminian: I will make a note to set some tests up in the near future. In the meantime, let's prioritize the ability to run some scripts on the dataset we are importing

hagoptaminian:
@pyphilia thanks for the detailed comments! basic dataset import/copy functionality is now working. if you think it's merge-worthy, let's merge this PR, and next we'll work on the functionality to run a script on the dataset. otherwise, we can keep working on this branch. let me know what you prefer, and in all cases if you have any comments on this latest code.

public/app/config/channels.js
```diff
@@ -1,3 +1,9 @@
 // this file needs to use module.exports as it is used both by react and
 // electron make sure this file is identical in both src/config/channels.js
 // and public/app/config/channels.js
+
+module.exports = {
+  SHOW_LOAD_DATASET_PROMPT_CHANNEL: 'prompt:dataset:load:show',
+  RESPOND_LOAD_DATASET_PROMPT_CHANNEL: 'prompt:dataset:load:respond',
+  CREATE_FILE_COPY: 'file:import',
```
pyphilia, line 8: Maybe I'll go with `LOAD_DATASET: 'dataset:load'` since we will have many operations on datasets later. Plus it goes better with all your other channels!

public/app/listeners/showLoadDatasetPrompt.js
```diff
@@ -0,0 +1,13 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+const { dialog } = require('electron');
+const { RESPOND_LOAD_DATASET_PROMPT_CHANNEL } = require('../config/channels');
+const logger = require('../logger');
+
+const showLoadDatasetPrompt = (mainWindow) => (event, options) => {
```
pyphilia, line 6: Here I would not have `options` as an argument, but as 
`const options = {
      filters: [{ name: 'json', extensions: ['json'] }],
    }; `

public/electron.js
```diff
@@ -241,11 +244,13 @@ ipcMain.on('load-page', (event, arg) => {
 });
 
 ipcMain.on(CREATE_FILE_COPY, (event, args) => {
```
pyphilia, line 246: Let's create a listener `loadDataset.js` with this function

pyphilia (changes requested):
I commented a bit more regarding `options`, after these changes we are good to go!

public/app/listeners/showLoadDatasetPrompt.js
hagoptaminian: 👍 is this because we won't be uploading other types of files?

PR #19, opened 2020-09-14
10/execute python

hagoptaminian:


public/app/config/channels.js
```diff
@@ -6,4 +6,5 @@ module.exports = {
   SHOW_LOAD_DATASET_PROMPT_CHANNEL: 'prompt:dataset:load:show',
   RESPOND_LOAD_DATASET_PROMPT_CHANNEL: 'prompt:dataset:load:respond',
   LOAD_DATASET: 'dataset:load',
+  EXECUTE_PYTHON_SCRIPT: 'script:python',
```
pyphilia, line 9: I would go for `script:python:execute` as we are going to have other operations such as kill, pause, etc.

public/app/listeners/executePythonScript.js
```diff
@@ -0,0 +1,22 @@
+const path = require('path');
+const { spawn } = require('child_process');
+const logger = require('../logger');
+
+const executePythonScript = (event, args) => {
+  const { scriptId } = args;
+  logger.debug(`executing script with scriptId ${scriptId}`);
+  logger.debug('expecting a file named copy_of_data.json to exist');
+  logger.debug(__dirname);
```
pyphilia, line 9: Is this log necessary?

src/components/Scripts.js
```diff
@@ -0,0 +1,32 @@
...
+import Typography from '@material-ui/core/Typography';
+import Button from '@material-ui/core/Button';
+import Main from './common/Main';
+import { EXECUTE_PYTHON_SCRIPT } from '../config/channels';
+
+const Scripts = () => {
+  const { t } = useTranslation();
+
+  const executePythonScript = () => {
+    window.ipcRenderer.send(EXECUTE_PYTHON_SCRIPT, { scriptId: 1 });
```
pyphilia, line 12: Maybe using `id` instead of `scriptId` is enough.

public/python.py
```diff
@@ -0,0 +1,10 @@
+import sys
+import json
+
+# expects sys.argv[1] to be path for json file, formatted as per graasp research
+with open(sys.argv[1]) as json_file:
+    allData = json.load(json_file)
+    with open("anonymized_actions.json", "w+") as anonymizedFile:
```
pyphilia, line 7: Where does this file comes from? 
I think to be less dependent on scripts, assure saving the files each time, and be able to list the resulting file in the database later, it will be necessary to save the file in our back-end (instead of in the script). If possible, can you return/print the result content in the python script, and get it back in the js backend to save it locally in `VAR_FOLDER`?

pyphilia (changes requested):
Nice! I let you some comments, the most important being saving the file using the javascript back-end instead of though the python files. We should not forget that these files will later be defined by the users.

public/app/config/channels.js
```diff
@@ -6,4 +6,5 @@ module.exports = {
   SHOW_LOAD_DATASET_PROMPT_CHANNEL: 'prompt:dataset:load:show',
   RESPOND_LOAD_DATASET_PROMPT_CHANNEL: 'prompt:dataset:load:respond',
   LOAD_DATASET: 'dataset:load',
+  EXECUTE_PYTHON_ALGORITHM: 'algorithm:python',
```
pyphilia, line 9: Could you change it to `algorithm:python:execute`?

public/app/listeners/executePythonScript.js
hagoptaminian: Nope, just kept it there for debugging. I will remove it.

public/python.py
hagoptaminian: @pyphilia I will create a separate issue for this - would be great if you can show how it's done.

Issue #15, opened 2020-09-10
Display a list of datasets


pyphilia:
Some example layout
![image](https://user-images.githubusercontent.com/11229627/92700168-cb2dda80-f34e-11ea-982b-43ca6db64b50.png)

Painguin:
close #20

PR #20, opened 2020-09-15
15/display datasets

Painguin:
Issue #15: Displays a list of datasets (from fake datasets for now).

### DIFF
--- a/src/components/Datasets.js
+++ b/src/components/Datasets.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { withRouter } from 'react-router';
+import { useTranslation } from 'react-i18next';
 import { makeStyles } from '@material-ui/core/styles';
 
 import Container from '@material-ui/core/Container';
@@ -88,6 +89,7 @@ const useStyles = makeStyles({
 
 const Datasets = () => {
   const classes = useStyles();
+  const { t } = useTranslation();
 
   // name column sorted ascending by default
   const [isAsc, setIsAsc] = React.useState(true);
@@ -141,7 +143,7 @@ const Datasets = () => {
                     handleSortColumn(DATASETS_TABLE_COLUMNS.NAME);
                   }}
                 >
-                  <strong>Name</strong>
+                  <strong>{t('Name')}</strong>
                 </TableSortLabel>
               </TableCell>
               <TableCell align="right">
@@ -211,15 +213,13 @@ const Datasets = () => {
                     >
                       <PublishIcon />
                     </IconButton>
-                    {!dataset.anonymized ? (
+                    {!dataset.anonymized && (
                       <IconButton
                         aria-label="anonymize"
                         onClick={() => handleAnonymize(dataset)}
                       >
                         <EqualizerIcon />
                       </IconButton>
-                    ) : (
-                      ''
                     )}
                     <IconButton
                       aria-label="edit"
--- a/src/components/Datasets.js
+++ b/src/components/Datasets.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { withRouter } from 'react-router';
+import { darken, makeStyles } from '@material-ui/core/styles';
 
 import Container from '@material-ui/core/Container';
 import Table from '@material-ui/core/Table';
@@ -13,7 +14,9 @@ import PublishIcon from '@material-ui/icons/Publish';
 import EqualizerIcon from '@material-ui/icons/Equalizer';
 import EditIcon from '@material-ui/icons/Edit';
 import DeleteIcon from '@material-ui/icons/Delete';
+import AddIcon from '@material-ui/icons/Add';
 
+import theme from '../theme';
 import Main from './common/Main';
 
 const fakeDatasets = [
@@ -69,7 +72,22 @@ function stableSort(array, comparator) {
   return stabilizedThis.map((el) => el[0]);
 }
 
+const useStyles = makeStyles({
+  addButton: {
+    color: theme.palette.secondary.main,
+    backgroundColor: theme.palette.primary.main,
+    position: 'absolute',
+    right: '16px',
+    bottom: '16px',
+    '&:hover, &.Mui-focusVisible': {
+      backgroundColor: darken(theme.palette.primary.main, 0.2),
+    },
+  },
+});
+
 const Datasets = () => {
+  const classes = useStyles();
+
   // name column sorted ascending by default
   const [isAsc, setIsAsc] = React.useState(true);
   const [orderBy, setOrderBy] = React.useState('name');
@@ -88,6 +106,10 @@ const Datasets = () => {
 
   const handleDelete = () => {
     // TODO: implement delete functionality
+  };
+
+  const handleAdd = () => {
+    // TODO: implement add functionality
   };
 
   const handleSortColumn = (column) => {
@@ -194,6 +216,14 @@ const Datasets = () => {
             )}
           </TableBody>
         </Table>
+        <IconButton
+          variant="contained"
+          aria-label="add"
+          className={classes.addButton}
+          onClick={() => handleAdd()}
+        >
+          <AddIcon />
+        </IconButton>
       </Container>
     </Main>
   );
--- a/src/components/Datasets.js
+++ b/src/components/Datasets.js
@@ -21,7 +21,6 @@ import AddIcon from '@material-ui/icons/Add';
 import theme from '../theme';
 import Main from './common/Main';
 import { DATASETS_TABLE_COLUMNS, ORDER_BY } from '../config/constants';
-import { fakeDatasets } from './exampleDatasets';
 import { sortByKey } from '../utils/sorting';
 
 const useStyles = makeStyles({
@@ -39,6 +38,34 @@ const useStyles = makeStyles({
     fontWeight: 'bold',
   },
 });
+
+// fake dataset that will be deleted later
+const fakeDatasets = [
+  {
+    name: 'Atomic Structure',
+    description: 'Atomic Structure Dataset of students during two months',
+    size: 413,
+    created: '7/13/20',
+    lastModified: '7/13/20',
+    anonymized: false,
+  },
+  {
+    name: 'Animals of the world',
+    description: 'Collection of all the referenced animals of world.',
+    size: 409600,
+    created: '7/12/18',
+    lastModified: '9/14/20',
+    anonymized: false,
+  },
+  {
+    name: 'Anonymized Atomic Structure',
+    description: 'Anonymized version of Atomic Structure Dataset',
+    size: 211,
+    created: '7/02/20',
+    lastModified: '7/02/20',
+    anonymized: true,
+  },
+];
 
 const Datasets = () => {
   const classes = useStyles();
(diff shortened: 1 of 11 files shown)


src/components/Datasets.js
```diff
@@ -0,0 +1,232 @@
...
+    name: 'Anonymized Atomic Structure',
+    description: 'Anonymized version of Atomic Structure Dataset',
+    size: 211,
+    created: '7/02/20',
+    lastModified: '7/02/20',
+    anonymized: true,
+  },
+];
+
+function descendingComparator(a, b, orderBy) {
```
pyphilia, line 49: In ES6, we can use arrow functions https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions It changes a how `this` is handled, but in this case it is not very important, but you might change it.

src/components/Datasets.js
```diff
@@ -0,0 +1,232 @@
...
+
+const useStyles = makeStyles({
+  addButton: {
+    color: theme.palette.secondary.main,
+    backgroundColor: theme.palette.primary.main,
+    position: 'absolute',
+    right: '16px',
+    bottom: '16px',
+    '&:hover, &.Mui-focusVisible': {
+      backgroundColor: darken(theme.palette.primary.main, 0.2),
```
pyphilia, line 83: Why a darker color? I think we can just use `theme.palette.primary.main`

src/components/Datasets.js
```diff
@@ -0,0 +1,232 @@
...
+                </TableRow>
+              ),
+            )}
+          </TableBody>
+        </Table>
+        <IconButton
+          variant="contained"
+          aria-label="add"
+          className={classes.addButton}
+          onClick={() => handleAdd()}
```
pyphilia, line 223: This can be simplify to `onClick={handleAdd}`

src/components/Datasets.js
```diff
@@ -0,0 +1,232 @@
...
+      <Container>
+        <h1>Datasets</h1>
+        <Table aria-label="simple table">
+          <TableHead>
+            <TableRow>
+              <TableCell align="left">
+                <TableSortLabel
+                  active={orderBy === 'name'}
+                  direction={orderBy === 'name' && !isAsc ? 'desc' : 'asc'}
+                  onClick={() => handleSortColumn('name')}
```
pyphilia, line 135: Let's refactor the column names as constants and saved in `src/config/constants`. Maybe have something similar to 
```
DATASETS_TABLE_COLUMNS = {
  NAME: 'name',
  SIZE: 'size'
}
```
etc..
You can check this example here https://github.com/graasp/graasp-desktop/blob/7cb48aa56424ed8b940770c7ca2c68532ebcb903/src/components/classrooms/StudentsTable.js . You might not need to make everything dynamic, but let's avoid writing repetitive strings.

src/components/Datasets.js
```diff
@@ -0,0 +1,232 @@
...
+    <Main content>
+      <Container>
+        <h1>Datasets</h1>
+        <Table aria-label="simple table">
+          <TableHead>
+            <TableRow>
+              <TableCell align="left">
+                <TableSortLabel
+                  active={orderBy === 'name'}
+                  direction={orderBy === 'name' && !isAsc ? 'desc' : 'asc'}
```
pyphilia, line 134: Refactor 'desc' and 'asc' as `ORDER_BY.DESC` and `ORDER_BY.ASC`

pyphilia (changes requested):
Really good first commit! I left some minor comments, let me know if something is not clear.

Painguin:
Thank you for your relevant comments @pyphilia! I'll make the changes.

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+                    </IconButton>
+                    {!dataset.anonymized ? (
+                      <IconButton
+                        aria-label="anonymize"
+                        onClick={() => handleAnonymize(dataset)}
+                      >
+                        <EqualizerIcon />
+                      </IconButton>
+                    ) : (
+                      ''
```
pyphilia, line 222: You can simplify with `!dataset.anonymized && <Icon>...`

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+                  direction={
+                    orderBy === DATASETS_TABLE_COLUMNS.NAME && !isAsc
+                      ? ORDER_BY.DESC
+                      : ORDER_BY.ASC
+                  }
+                  onClick={() => {
+                    handleSortColumn(DATASETS_TABLE_COLUMNS.NAME);
+                  }}
+                >
+                  <strong>Name</strong>
```
pyphilia, line 144: Translate

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+                  direction={
+                    orderBy === DATASETS_TABLE_COLUMNS.CREATED && !isAsc
+                      ? ORDER_BY.DESC
+                      : ORDER_BY.ASC
+                  }
+                  onClick={() => {
+                    handleSortColumn(DATASETS_TABLE_COLUMNS.CREATED);
+                  }}
+                >
+                  <strong>Created</strong>
```
pyphilia, line 174: Translate

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+                  direction={
+                    orderBy === DATASETS_TABLE_COLUMNS.LAST_MODIFIED && !isAsc
+                      ? ORDER_BY.DESC
+                      : ORDER_BY.ASC
+                  }
+                  onClick={() => {
+                    handleSortColumn(DATASETS_TABLE_COLUMNS.LAST_MODIFIED);
+                  }}
+                >
+                  <strong>Last Modified</strong>
```
pyphilia, line 189: Translate

pyphilia (commented):
Nice changes! I added some comments about translations, you add them in `src/langs/en.json` or `src/langs/fr.json`!

pyphilia:
It's all good for me, I let @hagoptaminian check and merge if possible.

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+import EqualizerIcon from '@material-ui/icons/Equalizer';
+import EditIcon from '@material-ui/icons/Edit';
+import DeleteIcon from '@material-ui/icons/Delete';
+import AddIcon from '@material-ui/icons/Add';
+
+import theme from '../theme';
+import Main from './common/Main';
+import { DATASETS_TABLE_COLUMNS, ORDER_BY } from '../config/constants';
+
+const fakeDatasets = [
```
hagoptaminian, line 24: @Painguin very nice PR and the component looks great. Here, move this `fakeDatasets` array to a file `exampleDatasets.js` and add that file to `.gitignore`, since we don't want to merge this code to master.

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+    name: 'Anonymized Atomic Structure',
+    description: 'Anonymized version of Atomic Structure Dataset',
+    size: 211,
+    created: '7/02/20',
+    lastModified: '7/02/20',
+    anonymized: true,
+  },
+];
+
+const descendingComparator = (a, b, orderBy) => {
```
hagoptaminian, line 51: move this function and the two below it to the `utils` folder under `src`, and perhaps add a comment above each one explaining how it is used and what for. (remember, we want to keep our components as clean as possible)

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+    },
+  },
+});
+
+const Datasets = () => {
+  const classes = useStyles();
+  const { t } = useTranslation();
+
+  // name column sorted ascending by default
+  const [isAsc, setIsAsc] = React.useState(true);
```
hagoptaminian, line 95: For consistency with other files, let's do: at the top of the file:

`import React, { useState } from 'react'`

and instead of `React.useState()` here just `useState()`

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+    } else {
+      setOrderBy(column);
+      setIsAsc(true);
+    }
+  };
+
+  return (
+    <Main content>
+      <Container>
+        <h1>Datasets</h1>
```
hagoptaminian, line 130: Wrap with `t('')` function!

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+        <Table aria-label="simple table">
```
hagoptaminian, line 131: let's make this `aria-label` descriptive of the actual table (e.g. 'table of datasets')

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+                  direction={
+                    orderBy === DATASETS_TABLE_COLUMNS.NAME && !isAsc
+                      ? ORDER_BY.DESC
+                      : ORDER_BY.ASC
+                  }
+                  onClick={() => {
+                    handleSortColumn(DATASETS_TABLE_COLUMNS.NAME);
+                  }}
+                >
+                  <strong>{t('Name')}</strong>
```
hagoptaminian, line 146: Two comments: 

1. In general, wrap any text-only components with the Material UI Typography component (https://material-ui.com/components/typography/#typography)
2. Instead of `<strong></strong>`, either (a) use the Typography component's `variant` property, (b) give the element styling via the `makeStyles` hook

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+              </TableCell>
+              <TableCell align="right" />
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            {stableSort(fakeDatasets, getComparator(isAsc, orderBy)).map(
+              (dataset) => (
+                <TableRow key={dataset.name}>
+                  <TableCell>
+                    <span style={{ fontSize: '20px' }}>{dataset.name}</span>
```
hagoptaminian, line 202: use `<Typography>` instead of `<span>`, and avoid `</br>` since it might render differently on different browsers. Also, strong preference is to declare any styles in the `makeStyles` hook rather than inline

src/components/Datasets.js
```diff
@@ -0,0 +1,255 @@
...
+};
+
+const getComparator = (isAsc, orderBy) => {
+  return isAsc
+    ? (a, b) => -descendingComparator(a, b, orderBy)
+    : (a, b) => descendingComparator(a, b, orderBy);
+};
+
+const stableSort = (array, comparator) => {
+  const stabilizedThis = array.map((el, index) => [el, index]);
```
hagoptaminian, line 68: (as mentioned above a short comment on what this function is doing would help). Perhaps instead of `stabilizedThis` use `stabilizedArray`

hagoptaminian (changes requested):
@Painguin very nice work! left some comments with requested changes

pyphilia:
The build fails because `exampleDatasets` is not included in the commits. Either include the file (and we will remove it once we will use the database data) or include the dataset constant in the file you need it.

hagoptaminian:
> The build fails because `exampleDatasets` is not included in the commits. Either include the file (and we will remove it once we will use the database data) or include the dataset constant in the file you need it.

@pyphilia you're right. @Painguin the PR changes look good - can you just move the `exampleDatasets` back into `Datasets.js` and add a comment above that these will be deleted later?

Painguin:
Thanks. I put it back into `Datasets.js`.

Issue #23, opened 2020-09-23
Add Results tab


pyphilia:
The results are the output dataset once a script is run against some dataset. 
This issue include adding the components related to the Results tab, as well as save the resulting dataset in a local `results` folder (well separated from `datasets`).
The Results tab should lead to a similar table as Datasets.

PR #31, opened 2020-09-28
27/handle db folders

### LINKED ISSUE #27: Handle db folders
On the application launch, create (if it does not exist) the needed folders (in the current case `datasets`).
On "Use Sample Database", remove corresponding files in the VAR_FOLDER.

Painguin:
close #27.

This PR also solves two newly discovered issues:
- The add button would not appear if no dataset was there.
- adding a dataset with the same name as another one would get overwrite it in the filesystem (Thus having two entries that link to the same dataset). The dataset name in the filesystem now corresponds to its id.

### DIFF
--- a/public/app/listeners/executePythonAlgorithm.js
+++ b/public/app/listeners/executePythonAlgorithm.js
@@ -34,11 +34,6 @@ const executePythonAlgorithm = (mainWindow, db) => (event, { datasetId }) => {
   process.on('close', (code) => {
     if (code !== 0) {
       logger.error(`python process exited with code ${code}`);
-
-      // delete tmp file
-      if (fs.existsSync(outputPath)) {
-        fs.unlinkSync(outputPath);
-      }
     } else {
       // save result in db
       const newDataset = createNewDataset({
@@ -47,10 +42,12 @@ const executePythonAlgorithm = (mainWindow, db) => (event, { datasetId }) => {
       });
       db.get('datasets').push(newDataset).write();
 
-      // delete tmp file
-      fs.unlinkSync(outputPath);
-
       logger.debug(`save resulting dataset at ${newDataset.filepath}`);
+    }
+
+    // delete tmp file
+    if (fs.existsSync(outputPath)) {
+      fs.unlinkSync(outputPath);
     }
   });
 };
--- a/public/app/listeners/executePythonAlgorithm.js
+++ b/public/app/listeners/executePythonAlgorithm.js
@@ -1,3 +1,4 @@
+const fs = require('fs');
 const path = require('path');
 const { spawn } = require('child_process');
 const ObjectId = require('bson-objectid');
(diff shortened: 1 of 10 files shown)


public/app/listeners/loadDataset.js
```diff
@@ -24,13 +23,16 @@ const createNewDataset = ({ id, name, filepath }) => {
 };
 
 const loadDataset = (mainWindow, db) => async (event, args) => {
-  const { fileLocation } = args;
+  const { fileLocation, id } = args;
```
pyphilia, line 26: I might be wrong, but in `LoadDataset.js` there is no `id` passed as argument. Maybe you can let the id be generated in `createNewDataset` and use the return value to create the `filepath` you need. Be careful because this function is also used in `executePythonAlgorithm.js`

pyphilia (changes requested):
Really nice catching these additional bugs!

public/app/listeners/loadDataset.js
Painguin: You are right. I missed its use in `executePythonAlgorithm.js` and messed up with the `id` logic... I'll make the changes the way you suggested.

public/app/listeners/executePythonAlgorithm.js
```diff
@@ -33,15 +34,23 @@ const executePythonAlgorithm = (mainWindow, db) => (event, { datasetId }) => {
   process.on('close', (code) => {
     if (code !== 0) {
       logger.error(`python process exited with code ${code}`);
+
+      // delete tmp file
+      if (fs.existsSync(outputPath)) {
```
pyphilia, line 39: What do you think about moving this conditional block after the if-else block? (and remove line 50-51)

pyphilia (approved):
Nice! I really like that idea of tmp file :P It is handled quite well in the load process.
I'll let @hagoptaminian have a look.

src/components/Datasets.js
```diff
@@ -188,7 +202,12 @@ class Datasets extends Component {
...
 
     if (!datasets.size) {
-      return <Main fullscreen>{t('No datasets available')}</Main>;
+      return (
+        <Main fullscreen>
+          {t('No datasets available')}
+          {this.renderAddButton()}
+        </Main>
+      );
     }
```
hagoptaminian, line 211: Why do we use these two conditionals (`!datasets` and `!datasets.size`)? Also, I noticed that while this component is loading, the 'No datasets available' text flashes briefly. This is not a huge deal but is not optimal. We should fix it. I can add it as a separate issue.

hagoptaminian (changes requested):
@Painguin very nice! this works perfectly. I have two main comments:

1. Having worked with these components today I feel we should use the naming `addDataset` rather than `createDataset`, and in the table and database `Added on` instead of `Created at`, etc. Technically we are not creating datasets here and the naming can become confusing for both us and users, especially when we will in fact _create_ datasets with our algorithms later. Please rename the functions/channels/constants in `loadDataset.js`, `Datasets.js`, etc. to become 'added' instead of 'created'.

2. The `Datasets.js` component has become quite big and maybe slightly hard to follow. We should work on breaking it up/refactoring into smaller pieces. I will add this as a separate issue.

Painguin:
As there is a conflict between naming "Added on" for loaded datasets and "Created at" for modified/processed datasets, we will handle the naming issue in issue #23 when both types of datasets will be separated.

hagoptaminian:
@Painguin @pyphilia very nice! Liking how the app is looking. I'll merge this to master.

hagoptaminian (approved):

PR #34, opened 2020-10-01
26/new load dataset

hagoptaminian:
@pyphilia @Painguin this PR adds tooltips to quick actions in `Datasets.js`, converts the 'load dataset' screen to a modal w/name and description fields, and cleans up some stylistic issues here and there.

@pyphilia as it stands, when one clicks on 'Add dataset' in the modal, one has to navigate away from `Datasets.js` and then back to see the component update. let's walk through an example where you show me how we can make this update automatic (via the redux setup). I've already made the component a class component to facilitate this.


public/app/listeners/loadDataset.js
```diff
@@ -24,29 +25,35 @@ const createNewDataset = ({ name, filepath }) => {
...
 
 const loadDataset = (mainWindow, db) => async (event, args) => {
-  const { fileLocation } = args;
-
-  // filename without extension
-  const filename = path
-    .basename(fileLocation)
-    .slice(0, -path.extname(fileLocation).length);
+  const { fileLocation, fileDescription } = args;
+  let { fileCustomName } = args;
```
pyphilia, line 37: I am not sure it is a good practice to edit `fileCustomName` of `args`.
Maybe consider 
```
const { fileLocation, ...., fileCustomName} = args;
const defaultFileName = path.basename.... 
// or without this tmp variable to avoid computing it if fileCustomName exists
const fileName = fileCustomName || defaultFileName

createNewDataset({
name: fileName
....
}) 
```

src/actions/dataset.js
```diff
@@ -44,6 +47,32 @@ export const getDatasets = () => (dispatch) => {
   }
 };
 
+export const addDataset = ({
```
pyphilia, line 50: `loadDataset` instead of `addDataset`

public/app/listeners/loadDataset.js
```diff
@@ -24,29 +25,35 @@ const createNewDataset = ({ name, filepath }) => {
...
 };
 
 const loadDataset = (mainWindow, db) => async (event, args) => {
-  const { fileLocation } = args;
-
-  // filename without extension
-  const filename = path
-    .basename(fileLocation)
-    .slice(0, -path.extname(fileLocation).length);
+  const { fileLocation, fileDescription } = args;
```
pyphilia, line 36: Just thinking, we might want some check on `fileLocation` such as whether the file really exists.

src/actions/dataset.js
```diff
@@ -44,6 +47,32 @@ export const getDatasets = () => (dispatch) => {
...
+  try {
+    window.ipcRenderer.send(LOAD_DATASET_CHANNEL, {
+      fileCustomName,
+      fileLocation,
+      fileDescription,
+    });
+    window.ipcRenderer.once(LOAD_DATASET_CHANNEL, async (event, dataset) => {
+      if (!dataset) {
+        toastr.error(ERROR_MESSAGE_HEADER, ERROR_ADDING_DATASET_MESSAGE);
+      } else {
```
pyphilia, line 64: You can add here a success `toastr` to give feedback to the user (even though the result is directly visible).

src/actions/dataset.js
```diff
@@ -44,6 +47,32 @@ export const getDatasets = () => (dispatch) => {
   }
 };
 
+export const addDataset = ({
+  fileCustomName,
+  fileLocation,
+  fileDescription,
+}) => (dispatch) => {
+  try {
```
pyphilia, line 55: You might want to add flags here to trigger the spinner to wait for the file to correctly load.

src/components/Datasets.js
```diff
@@ -135,65 +141,59 @@ class Datasets extends Component {
...
-              />
-            </IconButton>
-            <IconButton
-              aria-label="edit"
-              onClick={() => this.handleEdit(dataset)}
-            >
-              <EditIcon />
-            </IconButton>
-            {!dataset.anonymized && (
+            <Tooltip title="View dataset">
```
pyphilia, line 151: Translate your tooltips.

src/components/LoadDatasetButton.js
```diff
@@ -0,0 +1,57 @@
...
+      <Tooltip placement="left" title="Add dataset" arrow>
+        <IconButton
+          variant="contained"
+          className={classes.addButton}
+          onClick={handleClickOpen}
+        >
+          <AddIcon />
+        </IconButton>
+      </Tooltip>
+      <LoadDataset
```
pyphilia, line 47: Would this be clearer if it was called `LoadDatasetModal`?

src/components/LoadDataset.js
```diff
@@ -31,56 +79,126 @@ const LoadDataset = () => {
...
+        <DialogTitle id="create-new-item-form">
+          {t('Add new dataset')}
+        </DialogTitle>
+        <DialogContent className={classes.dialogContent}>
+          <div className={classes.addDataset}>
+            <TextField
+              autoFocus
+              margin="dense"
+              id="browse"
+              label="Select dataset"
```
pyphilia, line 121: Translate.

src/components/LoadDataset.js
```diff
@@ -31,56 +79,126 @@ const LoadDataset = () => {
...
+              className={classes.shortTextfield}
+              onChange={this.handleLocationInput}
+              value={fileLocation}
+            />
+            <IconButton onClick={this.handleBrowse}>
+              <SearchIcon />
```
pyphilia, line 127: I am not sure about this icon, for me it is mostly a `search` button and not really a `browse` button.

src/components/LoadDataset.js
```diff
@@ -31,56 +79,126 @@ const LoadDataset = () => {
...
+            </IconButton>
+          </div>
+
+          <TextField
+            margin="dense"
+            id="name"
+            label="Name"
```
pyphilia, line 134: Translate. I would maybe go for `Dataset Name` or something similar as `Name` is rather vague.

src/components/LoadDataset.js
```diff
@@ -31,56 +79,126 @@ const LoadDataset = () => {
...
+            value={fileCustomName}
+            onChange={this.handleCustomNameInput}
+            className={classes.shortTextfield}
+          />
+
+          <TextField
+            margin="dense"
+            id="description"
+            label="Description"
```
pyphilia, line 143: Translate.

src/config/messages.js
```diff
@@ -14,6 +14,7 @@ const ERROR_SETTING_DATABASE = 'There was an error setting the database';
 const ERROR_GETTING_DATABASE = 'There was an error retrieving the database';
 const ERROR_DELETING_DATASET_MESSAGE =
   'There was an error deleting the dataset';
+const ERROR_ADDING_DATASET_MESSAGE = 'There was an error adding the dataset';
```
pyphilia, line 17: Go with `ERROR_LOADING_DATASET_MESSAGE = 'There was an error loading the dataset';`

src/reducers/common.js
```diff
@@ -5,5 +5,9 @@ const updateActivityList = (flag) => {
   return (list) => list.pop();
 };
 
+const updateDatasets = (payload) => {
```
pyphilia, line 8: Since you are only pushing a dataset, maybe go with `pushDatasetToList` or something similar.

pyphilia (changes requested):
Nice work! Just a general thought about the modal. Would it be possible to highlight which field is optional and which are mandatory? It might help the user when he will fill in the modal. Plus, I think having everything full width would be better than half-half.

src/components/LoadDatasetButton.js
hagoptaminian: good call

public/app/listeners/loadDataset.js
hagoptaminian: nice! makes sense

public/app/listeners/loadDataset.js
hagoptaminian: Good call again!

src/components/LoadDataset.js
hagoptaminian: Let's leave this to resolve at a later time!

hagoptaminian:
Thanks @pyphilia very helpful comments! Pushed the changes

pyphilia (approved):

PR #37, opened 2020-10-12
feat: add results tab, save result dataset in results

### LINKED ISSUE #23: Add Results tab

pyphilia:
close #23


hagoptaminian (changes requested):
@pyphilia very nice and works perfectly! I left some comments, mainly around making the Results table look consistent with the Datasets table.

hagoptaminian (approved):

PR #43, opened 2020-10-15
35/change algo display

### LINKED ISSUE #35: change algorithms display
Handle the display of multiple algorithms. Display the algorithms in a table-like format.

Painguin:
This PR changes the `Algorithms` tab to display a table of the available algorithms and creates an `Executions` tab from which you can select and run an algorithm on some dataset.

The algorithms table features a delete button (functional) and an edit button (non-functional).
For now, two algorithms are available. They might not reflect exactly what we want but more and more specific ones will be added in the future.   
On application startup, the presence of the default `graasp` algorithms are checked and added if not present.
The database now contains an entry for the algorithms with the required information.

A table of executions is planned and will be done in an other PR.

close #35.


src/langs/fr.json
```diff
@@ -37,6 +37,15 @@
...
     "Description": "Description",
     "Cancel": "Annuler",
-    "Add dataset": "Ajouter le dataset"
+    "Add dataset": "Ajouter le dataset",
+    "Execute algorithm on a dataset": "Exécuter l'algorithme sur un dataset",
+    "Edit algorithm": "Éditer l'algorithme",
+    "Delete algorithm": "Supprimer l'algorithme",
+    "Dataset": "Dataset",
+    "Algorithm": "Algorithme",
+    "Load a dataset first": "Veuillez importer a dataset",
```
pyphilia, line 46: `un` instead of `a`

pyphilia (changes requested):
Very nice work! I like how you deal different programming languages.
You will surely need to rebase on my PR once it is merged.

src/components/Executions.js
```diff
@@ -0,0 +1,215 @@
...
+                variant="contained"
+                color="primary"
+                onClick={this.executeAlgorithm}
+              >
+                {t('Execute')}
+              </Button>
+            </Grid>
+          </Grid>
+        </Grid>
+      </Main>
```
hagoptaminian, line 187: The use of the Grid system here is really overkill. Simplify to:

```JavaScript
<div className={classes.container}>
  <FormControl variant="outlined" className={classes.formControl}>
    <Select id="datasetSelect">...</Select>
  </FormControl>
  <FormControl variant="outlined" className={classes.formControl}>
    <Select id="algorithmSelect">...</Select>
  </FormControl>
  <Button
    variant="contained"
    color="primary"
    onClick={this.executeAlgorithm}
    className={classes.button}
  >
    {t('Execute')}
  </Button>
</div>;
```

where:

```JavaScript
  container: {
    width: '50%',
    margin: '0 auto',
    marginTop: theme.spacing(2),
  },
  formControl: {
    marginTop: theme.spacing(2),
    width: '100%',
  },
  button: { 
    display: 'block', 
    margin: '0 auto', 
    marginTop: theme.spacing(2) 
  },
```

src/components/Executions.js
```diff
@@ -0,0 +1,215 @@
...
+        </Main>
+      );
+    }
+
+    return (
+      <Main>
+        <Grid container justify="center" className={classes.content}>
+          <Grid container justify="center" alignItems="center">
+            <Grid item className={classes.selectLabel}>
+              <Typography>{t('Dataset')}</Typography>
```
hagoptaminian, line 137: Instead of this, within `FormControl`, use `InputLabel` as shown in the example here https://material-ui.com/components/selects/#select

hagoptaminian: Or alternatively `FormHelperText`, also in that example

src/components/Executions.js
```diff
@@ -0,0 +1,215 @@
...
+            </Grid>
+            <Grid item xs={5}>
+              <FormControl variant="outlined" className={classes.formControl}>
+                <Select
+                  id="datasetSelect"
+                  value={datasetId || ''}
```
hagoptaminian, line 143: Will the part after the `||` ever be reached? If there are no datasets, we are rendering the Alert, right? Look into it - if redundant, better to remove it.

hagoptaminian (changes requested):
@Painguin Nicely done! Left some comments, mostly on style/layout, and ensuring consistent display across tables. Let me know if anything is unclear.

src/components/Algorithms.js
```diff
@@ -3,108 +3,231 @@ import { connect } from 'react-redux';
...
+              </IconButton>
+            </Tooltip>
+          </TableCell>
+        </TableRow>
+      );
     });
-  };
+  }
+
+  renderTable() {
```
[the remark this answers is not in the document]
Painguin: That's a valid point. I think refactoring into a reusable component is good idea. An interesting component we could also use are [datagrids](https://material-ui.com/components/data-grid/rows/#data-grid-rows).

public/algorithms/default_anonymization.py
```diff
@@ -0,0 +1,136 @@
+""" Anonymize a dataset: removes sensitive values and replaces names/ids by hashes """
```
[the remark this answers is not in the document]
Painguin: I agree this should be way we present algorithms. But I will leave that one for now just for the sake of having more than one algorithm. I will modify the presented algorithms later in an other PR.

Painguin:
Thank you for your reviews!
I have abstracted the tables into a reusable component. It requires two props: 
- `columns`: array of columns that contains information about each column:
  - `columnName`
  - `sortBy`: field on which we sort the column (`undefined` if not sortable)
  - `field`: field corresponding to the value to display
  - `alignColumn`: alignment of the head/column name
  - `alignField`: alignment of the value field
  - `bold`: whether the column name is bold
- `rows`: array of objects with each required field for each row

Let me know what you think! @pyphilia @hagoptaminian

src/components/Results.js
```diff
@@ -105,215 +90,156 @@ class Results extends Component {
...
+        columnName: t('Quick actions'),
+        field: 'quickActions',
+        alignColumn: 'right',
+        alignField: 'right',
+        bold: false,
+      },
+    ];
 
-    return sortedResults.map((result) => {
+    const rows = results.map((dataset) => {
```
pyphilia, line 160: I guess it should be called `result` instead of `dataset`.

src/components/Results.js
```diff
@@ -105,215 +90,156 @@ class Results extends Component {
...
-                <DeleteIcon />
-              </IconButton>
-            </Tooltip>
-          </TableCell>
-        </TableRow>
-      );
+      return {
+        key: id,
+        name,
+        algorithmId,
```
pyphilia, line 180: Could we display the algorithm's name here instead of the id?

pyphilia (changes requested):
I really like this refactor~ Here some minor comments.

pyphilia (changes requested):
I really like the refactor~ Here some comments.

hagoptaminian:
@Painguin this is looking great. love the latest additions. reviewing the files now, but in the meantime I have two comments:

1. make the 'Quick Actions' header consistent in font with the rest of the table headers
2. right now we are using two different Alert styles - in one case in the middle of the page, and in one case across the top. can we switch to using just one style? (up to you which)

hagoptaminian (approved):
One more comment in addition the above:

In Results.js, instead of 'Result from' being the ID of the Python algorithm, which has no meaning to and is never seen by the user, it's better to use the algorithm name

hagoptaminian:
@Painguin @pyphilia for me this looks good to merge

PR #48, opened 2020-10-20
feat: add basic information in dataset view tab

### LINKED ISSUE #38: Add information in Dataset View
- date range
- number of actions
- number of users
- ...

pyphilia:
This PR adds a small table with basic information in the dataset view page. 

close #38

### DIFF
--- a/src/components/common/JSONFileReader.js
+++ b/src/components/common/JSONFileReader.js
@@ -12,11 +12,13 @@ class JSONFileReader extends Component {
     content: PropTypes.string,
     size: PropTypes.number,
     t: PropTypes.func.isRequired,
+    collapsed: PropTypes.oneOfType([PropTypes.bool, PropTypes.number]),
   };
 
   static defaultProps = {
     content: Map(),
     size: 0,
+    collapsed: true,
   };
 
   state = {
@@ -44,10 +46,10 @@ class JSONFileReader extends Component {
   };
 
   render() {
-    const { content, size, t } = this.props;
+    const { content, size, t, collapsed } = this.props;
     const { json } = this.state;
 
-    if (!content) {
+    if (!content || !json) {
       return <Loader />;
     }
 
@@ -57,7 +59,7 @@ class JSONFileReader extends Component {
       );
     }
 
-    return <ReactJson collapsed src={json} />;
+    return <ReactJson collapsed={collapsed} src={json} />;
   }
 }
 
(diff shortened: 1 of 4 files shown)


Painguin (changes requested):
Looks very good! 
I was wondering if we could increase the `MAX_FILE_SIZE` since it is first all collapsed. I opened ~1MB dataset without any problem, the only issue arised when I tried unfolding the collapsed 100-action blocks (~2s delay). This can also be reduced by changing the `groupArraysAfterLength` parameter.

Also the back button can go very far down when unfolding the dataset. What do you think of putting it on the top right at the same level of the dataset name?

pyphilia:
Thanks @Painguin for your input! Indeed it does lag a bit I set `groupArraysAfterLength` to 50 which still lags but a bit less.

Painguin (approved):

hagoptaminian (commented):
@pyphilia nicely done! one thought: Right now, every time we click on the <> to view the dataset, it has to redo the calculations - these are a bit time-consuming (there's a lag on my computer of 5-10 seconds) and the screen is stuck while doing them. so what if we 'cached' the data by appending it to the DB? that way subsequent views are not as time-consuming. this can be done in a separate issue/PR, if you prefer.

a smaller comment: would be nice if #s that are >= 1000 are written as 1,000 (with a comma). in the dataset summary, this is often the case for actions. I think if say `const myNum = 27021` doing `myNum.toLocaleString('en')` does the trick

pyphilia:
@hagoptaminian Thanks Hagop for your comments! I will change the number format so that it is easier to read.

Concerning the cache, yes we could extract the data directly when loading them but then we would need to compute it each time it changes (the user edit the dataset). There are obviously pros and cons. But are you sure it is the computation which takes time? Isn't it just the file loading?

hagoptaminian:
@pyphilia to make sure it was the computation, I commented out the `<JSONFileReader />` component in `DataScreen.js`, and it still took those 5-10 seconds. Let me know if I'm missing something. 

And I think you're right, the challenge is if the user edits the file. I thought we could detect such changes and only re-run the computation when we find them, but I'm also thinking that the computation to detect changes will be time-consuming as well (though it could be a background task). And such a process can have other applications in our app (e.g. to detect when any dataset has been 'tampered with'). Let me know what you think!

In the meantime I think we can merge this PR

pyphilia:
Let's review with your computer what takes time, but obviously we will have to tune the desktop application to handle these cases since the researchers are bound to deal with huge datasets and complex/time-consuming algorithms.

PR #51, opened 2020-10-22
feat: add python algorithms

Painguin:
Adds python scripts. 

`Sanitize users` (alternative to `Hash users`) hashes the users and tries to generically find occurrences of usernames and userids with regular expressions and replaces them.
`2-anonimize geolocation` looks at the country, region and city and removes them (from city to country) if it doesn't match another user with the same value.
The other scripts (aside from `Hash users`) are just suppression scripts to remove fields that may be sensitive.


Painguin:
Since there are a lot of suppression files, one thing we could add later on is a generic suppression tool to allow the user to select the fields he wants to suppress (while we indicate/propose the potentially sensitive ones).

public/app/config/config.js
```diff
@@ -33,20 +33,80 @@ const PROGRAMMING_LANGUAGES = {
...
+    language: PROGRAMMING_LANGUAGES.PYTHON,
+  },
+  {
+    id: ObjectId().str,
+    name: 'Sanitize users',
     description:
-      'Hashes the user ids, replaces user/id occurences by the hash version, removes geolocation and data fields',
-    filename: 'default_anonymization.py',
-    filepath: path.join(ALGORITHMS_FOLDER, 'default_anonymization.py'),
+      'Replaces the occurances of usernames and user IDs (userid) by the hash of the userid',
```
pyphilia, line 47: occurrence

public/app/config/config.js
```diff
@@ -33,20 +33,80 @@ const PROGRAMMING_LANGUAGES = {
...
+    description: 'Suppress the data fields from a dataset',
+    filename: 'suppress_data.py',
+    filepath: path.join(ALGORITHMS_FOLDER, 'suppress_data.py'),
+    author: DEFAULT_AUTHOR,
+    language: PROGRAMMING_LANGUAGES.PYTHON,
+  },
+  {
+    id: ObjectId().str,
+    name: 'Suppress users',
+    description: 'Suppress the users field from a dataset',
```
pyphilia, line 74: Maybe `Suppress the field 'users' from a dataset` would be more clear?

public/app/config/config.js
```diff
@@ -33,20 +33,80 @@ const PROGRAMMING_LANGUAGES = {
...
+      'suppress_appInstanceResources_data.py',
+    ),
+    author: DEFAULT_AUTHOR,
+    language: PROGRAMMING_LANGUAGES.PYTHON,
+  },
+  {
+    id: ObjectId().str,
+    name: '2-Anonimize geolocation',
+    description:
+      'Mask country, region and city attributes when necessary so that no user can be uniquely identified',
```
pyphilia, line 107: Maybe have a less vague description and talk about the condition (if a user is alone it will be masked)

pyphilia (approved):
Nice! Let you some minor comments.

public/app/algorithms/hash_users.py
```diff
@@ -4,7 +4,11 @@
 # expects sys.argv[1] to be path for json file, formatted as per graasp research
 with open(sys.argv[1]) as json_file:
     allData = json.load(json_file)
-    with open(sys.argv[2], "w+") as anonymizedFile:
+    with open(sys.argv[2], 'w+') as anonymizedFile:
         for action in allData['data']['actions']:
-            action['user'] = hash(action['user'])
-        json.dump(allData, anonymizedFile)
+            if 'user' in action:
+                action['user'] = format(hash(action['user']), 'x')
```
hagoptaminian, line 10: by the way, check out https://stackoverflow.com/questions/27522626/hash-function-in-python-3-3-returns-different-results-between-sessions. It turns out that as of Python 3.3 the built-in `hash` function returns a different value for the same input whenever there is a new session. for various reasons this might be problematic in our app. not sure if we are able to load modules yet, but the recommended solution is to use the hashlib module

public/app/config/config.js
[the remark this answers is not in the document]
hagoptaminian: I see now that this scans the dataset and replaces user _name_ occurrences with a hash of the userId. makes sense! Maybe better phrasing: _Scan the dataset for occurrences of user names, and replace such occurrences with a hash of the corresponding user ID._

public/app/algorithms/suppress_data.py
```diff
@@ -0,0 +1,38 @@
...
+
+def main():
+    args = parse_arguments()
+
+    dataset = load_dataset(args.dataset_path)
+
+    actions = dataset['data']['actions']
+
+    for action in actions:
+        del action['data']
```
hagoptaminian, line 33: similar to above: what happens if an action doesn't have a `data` key?

public/app/algorithms/two_anonimize_geolocations.py
```diff
@@ -0,0 +1,89 @@
...
+
+    actions = dataset['data']['actions']
+
+    # get first geolocation for each user
+    geolocations = {}
+    for action in actions:
+        user = action['user']
+        if user not in geolocations:
+            geolocations[user] = action['geolocation']
+            geolocations[user]['range'] = [0,0]
```
hagoptaminian, line 37: so this file is a combination between k-anonymization and suppression (we are effectively deleting all these fields that we are emptying). maybe add to the algorithm description: _, with remaining geolocation information suppressed_

hagoptaminian (changes requested):
interesting and looks good. agree with you that parameterizing the suppression option makes sense. in general I think we need to find a clearer and consistent way of describing what each algorithm does, which exact keys it acts on. I will think about how. maybe we can include a quick action that opens a modal with an example of what the algorithm does.

public/app/algorithms/suppress_appInstanceResources_data.py
```diff
@@ -0,0 +1,38 @@
...
+
+def main():
+    args = parse_arguments()
+
+    dataset = load_dataset(args.dataset_path)
+
+    app_instance_resources = dataset['data']['appInstanceResources']
+
+    for app_instances_resource in app_instance_resources:
+        del app_instances_resource['data']
```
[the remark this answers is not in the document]
Painguin: An error happens... Checking if the key exists is the way to go :+1:

Painguin:
Thank you for your comments and suggestions!
I went through the algorithm descriptions and made them clearer.
Within the algorithms, I also made sure to check the presence of the field before accessing/deleting it... unless it is crucial for the algorithm in which case throwing an error makes sense.

hagoptaminian (approved):

PR #55, opened 2020-10-23
feat: adjust how dataset sizes are displayed in tables

hagoptaminian:
a small PR to adjust how dataset sizes are displayed in our tables. it now looks like this:

Screen Shot 2020-10-23 at 5 16 49 PM

### DIFF
--- a/src/utils/formatting.js
+++ b/src/utils/formatting.js
@@ -1,0 +1,14 @@
+// takes fileSize in KBs with 3 decimals places, and formats them
+// if size < 100, display size in KBs to 2dps, e.g. 19.432 --> '19.43KB'
+// if size <= 999.499, display size in KBs rounded to nearest integer, e.g. 431.855 --> '432KB'
+// if size > 999.499, display size in MBs to 2ps, e.g. 12521.482 --> 12.52MB'
+// eslint-disable-next-line import/prefer-default-export
+export const formatFileSize = (sizeInKb) => {
+  if (sizeInKb < 100) {
+    return `${sizeInKb.toFixed(2)}KB`;
+  }
+  if (sizeInKb <= 999.499) {
+    return `${Math.round(sizeInKb)}KB`;
+  }
+  return `${(Math.round(sizeInKb) / 1000).toFixed(2)}MB`;
+};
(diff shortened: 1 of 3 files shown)


src/utils/formatting.js
```diff
@@ -0,0 +1,14 @@
+// takes fileSize in KBs with 3 decimals places, and formats them
+// if size < 100, display size in KBs to 2dps, e.g. 19.432 --> '19.43KB'
+// if size <= 999.499, display size in KBs rounded to nearest integer, e.g. 431.855 --> '432KB'
+// if size > 999.499, display size in MBs to 2ps, e.g. 12521.482 --> 12.52MB'
+// eslint-disable-next-line import/prefer-default-export
+export const formatFileSize = (sizeInKb) => {
+  if (sizeInKb < 100) {
+    return `${sizeInKb.toFixed(2)}KB`;
+  }
+  if (sizeInKb <= 999.499) {
```
pyphilia, line 10: Why 999.499 in particular?

src/utils/formatting.js
```diff
@@ -0,0 +1,14 @@
+// takes fileSize in KBs with 3 decimals places, and formats them
+// if size < 100, display size in KBs to 2dps, e.g. 19.432 --> '19.43KB'
+// if size <= 999.499, display size in KBs rounded to nearest integer, e.g. 431.855 --> '432KB'
+// if size > 999.499, display size in MBs to 2ps, e.g. 12521.482 --> 12.52MB'
```
pyphilia, line 4: I would put these precise comments in the function to explain each `if` block, and have a more global description for the function here.

pyphilia (approved):
Really good! I left some minor comments.

src/utils/formatting.js
hagoptaminian: You're right, I should just use `< 999.5` instead (999.5 because anything >= than that should be displayed as 1.00MB).

src/utils/formatting.js
hagoptaminian: good call!

pyphilia (approved):

PR #56, opened 2020-10-28
feat: check python version on startup and add informative logo in Executions

### LINKED ISSUE #36: Check Python is installed
It might be a nice warning for the user to detect whether python is installed on the computer, and maybe which version.

This can be done by running a process executing `python --version`

Painguin:
close #36.


pyphilia (changes requested):
I see that you pushed the `public/shared` folder. Isn't it in `.gitignore`? If not could you add it? (and remove the folder from the pushed files).

Otherwise really good job! What do you think of this new architecture?

src/actions/settings.js
```diff
@@ -43,4 +53,23 @@ const setLanguage = async ({ lang }) => (dispatch) => {
...
+      CHECK_PYTHON_INSTALLATION_CHANNEL,
+      async (event, response) => {
+        dispatch(response);
+        dispatch(flagCheckingPythonVersion(false));
+      },
+    );
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err);
+    toastr.error(ERROR_MESSAGE_HEADER, ERROR_PYTHON_NOT_INSTALLED_MESSAGE);
```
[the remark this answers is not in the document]
Painguin: I am here displaying an error message and not dispatching an error for the `ToastrMiddleware`. This is the settings action file, maybe you are confounding with the listener file?

Painguin:
The toastr/shared files changes are great! It just took some time to get used to :sweat_smile:.
I noticed some types or messages disappeared with the changes, so I tried adding them back.

pyphilia (approved):

public/shared/messages.js
```diff
@@ -31,10 +31,30 @@ const ERROR_DELETING_DATASET_MESSAGE =
...
   'The result dataset was deleted successfully';
-  const SUCCESS_DELETING_ALGORITHM_MESSAGE =
+const SUCCESS_DELETING_ALGORITHM_MESSAGE =
   'The algorithm was deleted successfully';
 const ERROR_DELETING_RESULT_MESSAGE =
   'There was an error deleting the result dataset';
+const ERROR_PYTHON_NOT_INSTALLED_MESSAGE =
+  'Python is not installed on your system';
+const buildPythonWrongVersionMessage = (version) =>
+  `Your system's python version (${version}) is not supported`;
```
hagoptaminian, line 41: Use: _Your system is running Python version ${version}, which is not supported by this application. Please upgrade to Python 2.7.0 or better._

public/shared/messages.js
```diff
@@ -31,10 +31,30 @@ const ERROR_DELETING_DATASET_MESSAGE =
...
+const ERROR_CHECKING_PYTHON_INSTALLATION_MESSAGE =
+  'There was an error checking the installation of python';
```
hagoptaminian, line 43: Maybe: _There was an error verifying the installation of Python on your system. Try restarting the application. If the error persists, try reinstalling Python._

public/shared/messages.js
```diff
@@ -31,10 +31,30 @@ const ERROR_DELETING_DATASET_MESSAGE =
 const SUCCESS_DELETING_DATASET_MESSAGE = 'The dataset was deleted successfully';
 const SUCCESS_DELETING_RESULT_MESSAGE =
   'The result dataset was deleted successfully';
-  const SUCCESS_DELETING_ALGORITHM_MESSAGE =
+const SUCCESS_DELETING_ALGORITHM_MESSAGE =
   'The algorithm was deleted successfully';
 const ERROR_DELETING_RESULT_MESSAGE =
   'There was an error deleting the result dataset';
+const ERROR_PYTHON_NOT_INSTALLED_MESSAGE =
+  'Python is not installed on your system';
```
hagoptaminian, line 39: _Python was not found on your system. Please make sure you have Python 2.7.0 or better installed._

src/components/execution/PythonLogo.js
```diff
@@ -0,0 +1,94 @@
...
+  valid: {
+    fill: theme.palette.primary.main,
+  },
+  invalid: {
+    fill: 'gray',
+  },
+});
+
+const PythonLogo = ({ t, pythonVersion, classes }) => {
+  if (!pythonVersion) return null;
```
hagoptaminian, line 22: `pythonVersion` is the payload object from `checkPythonInstallation.js`, correct? If so, this line needs to be revised, as JS objects are never falsy.

src/components/execution/PythonLogo.js
```diff
@@ -0,0 +1,94 @@
...
+
+  const { valid, version } = pythonVersion;
+
+  return (
+    <Tooltip
+      title={`Python (${
+        version ? `${t('version')} ${version}` : t('not installed')
```
hagoptaminian, line 29: This may also need to be reconciled/made consistent with the `return null` above. If it is not installed, I expect we want to return `null`, right? In which case the condition after `:` should be something other than 'not installed'

src/components/execution/PythonLogo.js
hagoptaminian: ok, I see now that the initial state of `pythonVersion` in the reducer is `null`, so this line is for that condition. my mistake

src/components/execution/PythonLogo.js
hagoptaminian: Ok, now I see what's happening. Makes sense.

hagoptaminian (changes requested):
@Painguin very nice. the python logo in `Executions.js` looks slick. I left some small comments. one thought: What do you think of disabling the Execute button in `Executions.js` if Python is not found or is an unsupported version? Why allow executions to run if Python is not there/not supported?

Nice work!

Painguin:
Thank you! That's a good point. I have made that change in the newest commit.

hagoptaminian (changes requested):
@Painguin looks good! Two more small comments:

1. in `PythonLogo.js`, on line 29, do something like `version ? `${t('version')} ${version}` : t('was not found on your system. Please install Python 2.7.0 or better to run this application.')`
2. on the now-disabled 'Execute' button, consider adding a tooltip on hover saying the same ('Python was not found on your system. Please...')... if you think this is overkill (i.e. that users will be intuitively see the greyed out Python logo and hover over it), it's fine, no need to add it

hagoptaminian:
@Painguin while we're here please also go through the messages in `/public/shared/messages.js` and `/src/shared/messages.js` and make sure that all instances of 'python' are capitalized

hagoptaminian (approved):

pyphilia (approved):

Issue #58, opened 2020-11-02
show some kind of notification when an execution is running


hagoptaminian:
when you add a dataset and run a script on it in the Execute tab, there is no acknowledgment that the script is running. this is not an issue for small datasets, because the execution happens really fast and you get the toast notification that it executed successfully. but for medium and bigger datasets, it seems like nothing is happening. for example:

![ezgif-3-b43b178f6ec2](https://user-images.githubusercontent.com/19311953/97856019-f6ec9180-1cfb-11eb-9a83-97fdb302c9c6.gif)

one possible solution is converting the Execute button to a loading spinner while an execution is happening

PR #60, opened 2020-11-02
feat: add executions table

### LINKED ISSUE #44: Add executions table
In the `Executions` table, display a table of the executions. Failed, running and successful runs should be displayed.

pyphilia:
close #44

@hagoptaminian and I believe it also include #57 ?


Painguin (changes requested):
Looks very good @pyphilia! I like how you can click on the dataset or result in the executions tab and it opens the corresponding view :smile:. 
There were only some small issues when I tried to run multiple executions at the same time where only one of executions got its status set to a success, the others were left with a pending status for some reason (even though I believe they were all finished and actually successful, reloading the executions tab set them all correctly as successful). I don't know if this is an easy fix, I know that concurrency can be very tricky.

pyphilia:
@Painguin Thank you for your review! They were really good points, I indeed change the handling of executions so that the responses go to different channels `EXECUTE_ALGORITHM_CHANNEL_[executionId]` (a trick I learned for Graasp Desktop) ;)

public/app/listeners/executeAlgorithm.js
```diff
@@ -0,0 +1,140 @@
...
+      name: algorithmName,
+      language,
+    } = db.get(ALGORITHMS_COLLECTION).find({ id: algorithmId }).value();
+
+    const tmpPath = path.join(RESULTS_FOLDER, `tmp_${executionId}.json`);
+
+    // prepare success callback function
+    // copy tmp as new result dataset
+    // update execution and return it
+    const onSuccess = () => {
```
hagoptaminian, line 68: ok I've been trying to debug an issue but maybe it's more efficient to collaborate on it :)
There is something off I suspect in both or either this `onSuccess` callback and/or `process.on('close', ...)` in `executePythonAlgorithm`. specifically... try executing an algorithm as normal (just once). if you look at `db.json`, even if after the algorithm has executed successfully, its status remains `"pending"`. HOWEVER, if you run another algorithm, the **previous** algorithm becomes marked as `"status": "success"`. but as a result this is quite buggy because if you exit the app and reopen it then an algorithm previously showing as complete now shows as pending.
so something is off, my guess is in this function or in `executePythonAlgorithm`

hagoptaminian (changes requested):
@pyphilia very nice 👍 I left a bunch of comments, and I'll do another review after you look at them because I feel there's already a decent amount in there :) 

Otherwise, I thought one nice thing to add to the table is something like 'Execution Time', i.e. how long the execution took. some algos, like Sanitize users, took quite a while to run

let me know if anything is unclear!

public/app/config/config.js
```diff
@@ -113,6 +113,13 @@ const GRAASP_ALGORITHMS = [
 
 const ACCEPTED_PYTHON_VERSIONS = ['2.7', '3'];
 
+const EXECUTION_STATUSES = {
+  PENDING: 'pending',
+  RUNNING: 'running',
```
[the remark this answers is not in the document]
pyphilia: Pending is an execution which was not yet executed (and is not being executed) while running is currently being executed.

public/app/listeners/deleteExecution.js
```diff
@@ -0,0 +1,29 @@
...
+const { EXECUTIONS_COLLECTION } = require('../db');
+const { ERROR_GENERAL } = require('../../shared/errors');
+const {
+  DELETE_EXECUTION_SUCCESS,
+  DELETE_EXECUTION_ERROR,
+} = require('../../shared/types');
+
+const deleteExecution = (mainWindow, db) => async (event, { id }) => {
+  try {
+    db.get(EXECUTIONS_COLLECTION).remove({ id }).write();
```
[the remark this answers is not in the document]
pyphilia: Really good point. If you don't mind, I will open a new issue regarding this additionally to dealing with running processes/executions on app close.

public/app/listeners/createExecution.js
```diff
@@ -0,0 +1,37 @@
...
+  { algorithmId, datasetId },
+) => {
+  try {
+    // add execution
+    const execution = {
+      id: ObjectId().str,
+      algorithmId,
+      datasetId,
+      executedAt: Date.now(),
+      status: EXECUTION_STATUSES.PENDING,
```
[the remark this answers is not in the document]
pyphilia: Indeed nothing is displayed for pending executions. The whole goal of pending is to let the possibility (in the future) to start the execution at a later time point instead of running the execution right away. Maybe it will ease the process for pause/start a process (but I am not yet sure how it would work). So I was thinking about displaying a start button for pending executions, let me know what you think @hagoptaminian and @Painguin as well.

src/components/Executions.js
```diff
@@ -179,60 +370,69 @@ class Executions extends Component {
 
     return (
       <Main>
-        <div className={classes.container}>
-          <div className={classes.pythonLogo}>
-            <PythonLogo />
+        <Container>
+          <h1>{t('Executions')}</h1>
```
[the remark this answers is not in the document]
pyphilia: I prefer increasing the margin. JC gave us feedback the other day that it would be more consistent if we had an add button as for datasets and algorithms to create new executions. So this will eventually change later.

src/components/Executions.js
```diff
@@ -1,25 +1,40 @@
...
+import Container from '@material-ui/core/Container';
 import { withTranslation } from 'react-i18next';
 import MenuItem from '@material-ui/core/MenuItem';
 import FormControl from '@material-ui/core/FormControl';
 import InputLabel from '@material-ui/core/InputLabel';
+import SuccessIcon from '@material-ui/icons/Done';
+import ErrorIcon from '@material-ui/icons/Clear';
+import IconButton from '@material-ui/core/IconButton';
+import DeleteIcon from '@material-ui/icons/Delete';
+import RunningIcon from '@material-ui/icons/Refresh';
```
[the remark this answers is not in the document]
pyphilia: I love the circular with progress! But there is no way to determinate the progress value of processes yet.. so I went for the simple one for now.

public/app/listeners/executeAlgorithm.js
pyphilia: I am not sure about the bug you are experiencing. Is it related to the db from the developer tab? I got some inaccuracies with it when working on Graasp Desktop. We can go through it together someday, but my guess would go for an deprecated value at some time.

public/app/listeners/createExecution.js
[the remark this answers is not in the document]
Painguin: I think it is good to automatically start the execution as it is added, but we could have a `start now` checkbox if we want to allow starting the execution later. And then have a `start`, `restart`, `pause` or `unpause` button depending on the state of the execution.

public/app/listeners/executeAlgorithm.js
hagoptaminian: no, doesn't have to do with the db from the developer tab. it can be reproduced on ceaeccf exactly as I described in my comment. in any case, the changes you've made in `executeAlgorithm.js` in the last commit fix this error: specifically, you now set an execution as running (lines 51-54), and in the `onSuccess` callback you `.write()` the update to `EXECUTIONS_COLLECTION` instead of reading it (with `.value()`). maybe you caught/fixed the error independently of my comment... in any case, it is fixed

Painguin:
Running multiple executions works well now and I have no status issues (unless I spam the execute button very fast :laughing:), good job!
I noticed one issue, when I delete a result, the execution is still there and I can click on the (empty) `Result` field which would open the result view with an error (either showing a white screen or the last result shown). So we either remove the executions corresponding to the result as it is deleted or not allow clicking on the `Result` field.
(Not directly related to the PR) Also when there is an error when fetching the result, we should catch the error type in the reducer and empty the state of `current` so that it doesn't display a previously shown result.

public/app/listeners/deleteExecution.js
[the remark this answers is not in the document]
hagoptaminian: makes sense to open an issue to fix this. in the meantime, in `Executions.js`, let's disable the trash/delete button for executions that are running

public/app/listeners/createExecution.js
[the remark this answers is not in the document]
hagoptaminian: interesting idea. let's think about it. after clicking Execute, why would a user need to 're-confirm' the execution (hypothetically marked as pending)? but I am open to it. in any case since an execution now moves from pending to running (after the changes to `executeAlgorithm.js`), it's less of an issue...

src/components/Executions.js
```diff
@@ -244,20 +258,32 @@ class Executions extends Component {
...
+          {resultName}
         </Button>
       );
 
       let statusIcon = null;
       switch (status) {
         case EXECUTION_STATUSES.SUCCESS:
-          statusIcon = <SuccessIcon style={{ color: 'green' }} />;
+          statusIcon = (
+            <Tooltip title={t('Successful execution')}>
```
hagoptaminian, line 269: Use 'Execution successful'

src/components/Executions.js
```diff
@@ -244,20 +258,32 @@ class Executions extends Component {
...
+              <SuccessIcon style={{ color: 'green' }} />
+            </Tooltip>
+          );
           break;
         case EXECUTION_STATUSES.ERROR:
-          statusIcon = <ErrorIcon style={{ color: 'red' }} />;
+          statusIcon = (
+            <Tooltip title={t('An error occured')}>
```
hagoptaminian, line 276: 'An error occurred during execution'

src/components/Executions.js
```diff
@@ -244,20 +258,32 @@ class Executions extends Component {
...
+              <ErrorIcon style={{ color: 'red' }} />
+            </Tooltip>
+          );
           break;
         case EXECUTION_STATUSES.RUNNING:
-          statusIcon = <RunningIcon style={{ color: 'orange' }} />;
+          statusIcon = (
+            <Tooltip title={t('The execution is running...')}>
```
hagoptaminian, line 283: 'Execution running...'

src/components/Executions.js
```diff
@@ -244,20 +258,32 @@ class Executions extends Component {
...
+              <CircularProgress />
```
hagoptaminian, line 284: add `size={26}` (or thereabouts, 26 looked good to me). but the default is definitely a bit big

hagoptaminian (changes requested):
thanks @pyphilia! this one seems to be hectic 😅 I am still experiencing a few bugs... specifically:

- execution Result names are lost between sessions/app refreshes:

ezgif-1-b584cdcd6193

- wrong behavior when deleting a previously complete execution while another one is running:

ezgif-1-4d0cba085897


- when an execution is complete and you delete the original dataset, no dataset name shows up (this should be stored with the execution):

ezgif-1-76844929f025

- also: since we are now displaying execution history in the Executions tab, in `Executions.js`, lines 370 and 380 (`if (!datasets.size)` and `if (!algorithms.size)`), instead of showing the Alert, we should show the page as normal, but have the Execute button disabled. If I've run a bunch of executions on a dataset, but then deleted the original dataset, it doesn't make sense for me not to be able to see the execution history

- in general, I think it's a good idea for you to test/play around with the application given this new code... add different sizes of datasets, run different algorithms, navigate between tabs, close the application and restart it, try different 'states' (e.g. all datasets deleted)... so that you can also catch some of these bugs...

public/app/listeners/createExecution.js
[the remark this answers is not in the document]
pyphilia: I'll make an issue with it. We talk about it tomorrow.

pyphilia:
I resolved the bugs mentioned. It might be worth writing some tests before merging, so I will wait on the datasets and algorithms tests before going on.

hagoptaminian (approved):
nice one @pyphilia 👍

hagoptaminian:
oh and this does indeed close #58 (#57 also but that was a duplicate which I deleted) @pyphilia

pyphilia:
This PR is back! A big rebase was done and I added end-to-end tests with the noticed bugs to be sure they are not there anymore. 
What this PR contains:
- **Results are now saved in Datasets** : it became really cumbersome to deal with either datasets or results because at each step we would need to search in datasets or in results for the actual source given the id... In order to differentiate between a "dataset" and a "result" is their type: either `source` or `result`.
- At each execution, a row is added in the table. The execution is executed immediately. Corresponding toastr are displayed for each execution even if the text is similar (override `preventDuplicates` of Toastr)
- An execution can be `pending`, `running`, `success`or `error`. An execution can be deleted or canceled (stop during execution).
- If the user quits the app before some executions end, these are killed and set as `error`.

src/components/common/Main.js
```diff
@@ -62,6 +62,7 @@ class Main extends Component {
     }).isRequired,
     children: PropTypes.node.isRequired,
     fullScreen: PropTypes.bool,
+    id: PropTypes.string.isRequired,
```
Painguin, line 65: We don't set the id everytime, we should remove the `.isRequired` and set some defaultProp (`undefined` or `''`)

src/components/execution/AddExecutionForm.js
```diff
@@ -17,17 +17,29 @@ import Loader from '../common/Loader';
...
+  EXECUTIONS_ALERT_NO_DATASET_ID,
+  EXECUTIONS_ALGORITHMS_SELECT_ID,
+  EXECUTIONS_DATASETS_SELECT_ID,
+  EXECUTIONS_EXECUTE_BUTTON_ID,
+} from '../../config/selectors';
 
 const styles = (theme) => ({
+  wrapper: {
+    width: '50%',
+    margin: theme.spacing(2, 'auto', 0),
```
Painguin, line 41: Didn't know we could to that... Should be useful :slightly_smiling_face:

src/langs/fr.json
```diff
@@ -93,6 +93,17 @@
...
     "User": "Utilisateur",
-    "You can use the 'utils' file to write functions you want to use across your custom algorithms": "Vous pouvez utiliser le fichier 'utils' pour écrire des fonctions que vous voulez utiliser à travers vos algorithmes personnalisés"
+    "You can use the 'utils' file to write functions you want to use across your custom algorithms": "Vous pouvez utiliser le fichier 'utils' pour écrire des fonctions que vous voulez utiliser à travers vos algorithmes personnalisés",
+    "From Algorithm": "De l'Algorithme",
+    "An error occurred during execution": "Une erreur est survenue durant l'exécution",
+    "Execution successful": "Execution réussie",
+    "Execution running...": "Execution en cours...",
+    "Remove execution": "Effacer l'exécution",
+    "Status": "Statut",
+    "Executed": "Executé",
```
Painguin, line 103: Exécution et Exécuté

src/shared/errors.js
```diff
@@ -6,11 +6,15 @@ const ERROR_MISSING_FILE = 'ERROR_FILE_DOES_NOT_EXIST';
...
+const UNKNOWN_PROGRAMMING_LANGUAGE = 'UNKNOWN_PROGRAMMING_LANGUAGE';
 
 module.exports = {
   ERROR_GENERAL,
   ERROR_MISSING_FILE,
   ERROR_PYTHON_PROCESS,
   ERROR_PYTHON_UNSUPPORTED_VERSION,
   ERROR_PYTHON_NOT_INSTALLED,
+  ERROR_EXECUTION_PROCESS,
+  UNKNOWN_PROGRAMMING_LANGUAGE,
```
Painguin, line 19: Maybe rename to `ERROR_UNKNOWN_PROGRAMMING_LANGUAGE` so that all errors start with "ERROR"

src/shared/messages.js
```diff
@@ -20,12 +22,10 @@ const ERROR_GETTING_DATABASE_MESSAGE =
...
-const buildPythonProcessErrorMessage = (code) =>
-  `The Python process exited with code ${code}`;
-const ERROR_EXECUTING_PYTHON_ALGORITHM_MESSAGE =
-  'There was an error executing the Python algorithm';
-const SUCCESS_EXECUTING_PYTHON_ALGORITHM_MESSAGE =
-  'The Python algorithm was executed successfully';
+const buildProcessErrorMessage = (code) =>
+  `The process exited with code ${code}`;
+const SUCCESS_EXECUTING_ALGORITHM_MESSAGE =
+  'The execution was executed successfully';
```
Painguin, line 28: Now that there is `buidExecutingAlgorithmSuccessMessage`, is `SUCCESS_EXECUTING_ALGORITHM_MESSAGE` still needed?

src/shared/messages.js
```diff
@@ -79,6 +79,21 @@ const ERROR_GETTING_UTILS_MESSAGE =
...
 const SUCCESS_SAVING_UTILS_MESSAGE = 'The utils file was saved successfully';
+const ERROR_CREATING_EXECUTION_MESSAGE =
+  'There was an error creating the execution';
+const ERROR_GETTING_EXECUTIONS_MESSAGE =
+  'There was an error retrieving the executions';
+const ERROR_DELETING_EXECUTION_MESSAGE =
+  'There was an error deleting the execution';
+const SUCCESS_DELETING_EXECUTION_MESSAGE =
+  'The execution was deleted successfully';
+const buidExecutingAlgorithmSuccessMessage = (name) => {
```
Painguin, line 90: ~buid~ -->  build

test/commands.js
```diff
@@ -0,0 +1,48 @@
...
+import { expect } from 'chai';
+import {
+  EXECUTIONS_MENU_ITEM_ID,
+  QUIT_MENU_ITEM_ID,
+  RESULTS_MENU_ITEM_ID,
+} from '../src/config/selectors';
+import { openDrawer } from './utils';
+
+// eslint-disable-next-line import/prefer-default-export
+export const setUpClientCommandes = ({ client }) => {
```
Painguin, line 11: `setUpClientCommands`

Painguin (changes requested):
Good to see this PR back! I'm happy with how it turned out, good job @pyphilia! I left some small comments.

also: add the type to `my dataset` in the sample database. It doesn't show up otherwise.

public/app/listeners/deleteExecution.js
```diff
@@ -0,0 +1,31 @@
...
+
+    // todo: kill process
```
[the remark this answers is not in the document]
pyphilia: Indeed, forgot this one 😅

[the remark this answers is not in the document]
pyphilia: In fact it is not necessary as it `deleteExecution` happens when the execution had ended.

test/executions/executions.test.js
```diff
@@ -430,7 +430,7 @@ describe('Executions Scenarios', function () {
         );
       });
       describe('Two applications', () => {
-        it(
+        it.only(
```
Painguin, line 433: remove the `.only`

Painguin (approved):

public/app/listeners/deleteExecution.js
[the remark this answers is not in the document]
Painguin: Ah, makes sense

hagoptaminian (commented):
@pyphilia this looks really really good. a few small comments:

- not sure if this is expected behavior, but when you delete a dataset on which an execution has been performed, if you go back to the executions table you see that the algorithm which was run on the now 'unknown' dataset is also marked 'unknown'

- when the execution status is ❌, the tooltip says 'An error occurred during execution'. but we also get the ❌ when we stop the execution. so maybe the tooltip can say: 'An error occurred or the execution was stopped by the user' 

- I think it would be useful if in a result's dataset view we had an export button as well... because quite often you go to view the result but get 'This file is too big to be displayed.' and then you have to intuit that you can export (and therefore view) the dataset from the Results page. If feasible and you think it makes sense, would add to this PR (even if the export button also shows up on the view page of 'original' datasets)

pyphilia:
Thanks for the feedback @hagoptaminian !

- We discussed a bit about it. I can think of a solution where the execution contains the id of its corresponding resource or their name if they are deleted. It creates a bit of callback when we delete stuff, but I don't think there is a better solution to avoid "Unknown" resources.
- I agree, I will change it
- I will create a new issue with this.

PR #61, opened 2020-11-02
feat: edit dataset from interface

### LINKED ISSUE #41: Edit dataset from Dataset view's React-JSON Component

pyphilia:
close #41


public/app/listeners/setDatasetFile.js
```diff
@@ -0,0 +1,37 @@
...
+  try {
+    const { filepath } = db.get(DATASETS_COLLECTION).find({ id }).value();
+
+    if (!fs.existsSync(filepath)) {
+      return mainWindow.webContents.send(
+        SET_DATASET_FILE_CHANNEL,
+        ERROR_MISSING_FILE,
+      );
+    }
+    const contentAsString = JSON.stringify(content);
```
Painguin, line 21: This sets all the json content of the file into one line which is not visually pleasant when looking at the file itself. Use `JSON.stringify(content, null, 2)` instead to have indentation and all.

src/components/dataset/DatasetScreen.js
```diff
@@ -86,10 +86,13 @@ class DatasetScreen extends Component {
...
-    const content = dataset?.get('content') || '';
-    const parsedContent = JSON.parse(content)?.data;
+    const content = dataset?.get('content');
+    if (!content) {
+      return null;
+    }
+    const parsedContent = JSON.parse(content)?.data || {};
 
-    const { actions = [] } = parsedContent;
+    const actions = parsedContent?.actions.filter((action) => action) || [];
```
Painguin, line 95: If the actions are deleted, then there is an error on the `filter` call. Should be `parsedContent?.actions?.filter`

Painguin (changes requested):
Good feature! 

Each time I do some modification, the editor view is reset and everything is collapsed again. This can be very annoying when doing multiple small modifications. Is it possible not to reset the view? Maybe we still save the file after each change but we don't update the content with the reducer? I don't know if the JSON editor keeps the modifications or not in that case...

hagoptaminian (commented):
nice feature and works well. agree with @Painguin's comment that it would be nice if we could stay in the section of the file we are editing after confirming an edit (rather than having it refresh and collapse). a few other comments:

1. we access this edit view via the `View dataset` quick action - what are we expecting to do with the `Edit dataset` one?

2. currently this is only enabled for original datasets loaded into the app - i.e. not for results generated from executions. was this intentional?

3. would be nice if when attempting to delete a key in the json, we get a prompt confirming the delete/saying it is irreversible

pyphilia:
Thank you for your reviews!

@hagoptaminian 
1 - Honestly I am not sure. Maybe we should have a same quick action for both view and edit. What do you think?
2 - Yes, I guess results shouldn't be editable unless they are set as datasets. If we could edit a result, then it would mislead the user about the processing the algorithm applied, don't you think?
3 - It would be nice indeed, I will add it.

pyphilia:
The collapsed elements remain after an update. 
The strategy was to avoid refreshing `DatasetScreen.js`, because it would re-render `JSONFileEditor.js` and thus reset the collapsing. Therefore, I only use the necessary properties in `DatasetScreen.js`,  `JSONFileEditor.js` and `DatasetInformationTable.js` to avoid unnecessary rendering (particularly in the main screen). That way, the table is also updating alongside the json file.

hagoptaminian (commented):
looks good and works well! on our discussion above:

1. agreed that we should eliminate the Edit dataset icon (and have the current View dataset icon cover both)
2. good point. but I was thinking... imagine I ran a few algorithms on a dataset, and then I wanted to finalize the result, for example by deleting the appInstanceResources key in a dataset (because I don't need it, don't want to publish it)... it would be nice to be able to do this. of course, the user can still export the dataset and reload it via the Datasets tab, and edit it that way, but it's a bit of a longer process... so I'm leaning toward giving the users the ability to edit results as well.

public/app/listeners/setDatasetFile.js
```diff
@@ -0,0 +1,37 @@
...
+const setDatasetFile = (mainWindow, db) => async (e, { id, content }) => {
+  try {
+    const { filepath } = db.get(DATASETS_COLLECTION).find({ id }).value();
+
+    if (!fs.existsSync(filepath)) {
+      return mainWindow.webContents.send(
+        SET_DATASET_FILE_CHANNEL,
+        ERROR_MISSING_FILE,
+      );
+    }
```
Painguin, line 20: This is a better way of handling missing files than checking the `ENOENT` error code :+1: We should do that everywhere

public/shared/messages.js
```diff
@@ -62,6 +62,10 @@ const SUCCESS_EXPORTING_DATASET_MESSAGE =
 const ERROR_EXPORTING_RESULT_MESSAGE =
   'There was an error exporting the result';
 const SUCCESS_EXPORTING_RESULT_MESSAGE = 'The result was exported successfully';
+const ERROR_EDITING_DATASET_FILE_MESSAGE =
+  'There was an error editing the dataset file';
+const ERROR_SETTING_DATASET_FILE_MESSAGE =
+  'There was an error editing the dataset';
```
Painguin, line 68: `ERROR_EDITING_DATASET_FILE_MESSAGE` is not used anywhere. I assume it got replaced by `ERROR_SETTING_DATASET_FILE_MESSAGE` and should be deleted

src/components/common/JSONFileEditor.js
```diff
@@ -59,10 +72,30 @@ class JSONFileReader extends Component {
...
+      : {};
+
     return (
-      <ReactJson groupArraysAfterLength={50} collapsed={collapsed} src={json} />
+      <ReactJson
+        groupArraysAfterLength={50}
+        collapsed={collapsed}
+        src={json}
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        {...editProps}
```
Painguin, line 89: I didn't know you could do that :open_mouth:

src/shared/messages.js
```diff
@@ -62,6 +62,10 @@ const SUCCESS_EXPORTING_DATASET_MESSAGE =
 const ERROR_EXPORTING_RESULT_MESSAGE =
   'There was an error exporting the result';
 const SUCCESS_EXPORTING_RESULT_MESSAGE = 'The result was exported successfully';
+const ERROR_EDITING_DATASET_FILE_MESSAGE =
+  'There was an error editing the dataset file';
+const ERROR_SETTING_DATASET_FILE_MESSAGE =
+  'There was an error editing the dataset';
```
Painguin, line 68: delete `ERROR_EDITING_DATASET_FILE_MESSAGE`

Painguin (approved):
Good job @pyphilia! Specially on making the table reactive to the changes.

PR #62, opened 2020-11-02
39/export button

### LINKED ISSUE #39: Open Dataset and/or Result in folder
Some documentation https://github.com/electron/electron/blob/master/docs/api/shell.md

pyphilia:
This PR was initially to open files in folder, but I thought that it was more useful to go directly with an export button (and not expose our local storage because it could lead to errors afterwards). Let me know what you think.

close #39


Painguin (changes requested):
Looks good! Left some minor comments.

src/components/common/ExportButton.js
```diff
@@ -0,0 +1,69 @@
...
+
+  handleExport = () => {
+    const { dispatchExportFile, name, flagType, channel, id } = this.props;
+    dispatchExportFile({ name, flagType, channel, id });
+  };
+
+  render() {
+    const { tooltipText } = this.props;
+    return (
+      <Tooltip title={tooltipText} key="visualize">
```
hagoptaminian, line 54: I believe `key={visualize}` should be 'export'... but I was looking at https://material-ui.com/api/tooltip/ and it doesn't look like Tooltips take a key prop. so I'm not sure it is doing anything.

hagoptaminian (changes requested):
good idea to do it this way! looks great and works well.

was just wondering if it's worth having one generic listener (`exportFile.js`) rather than two separate ones for a result and a dataset (`exportDataset.js` and `exportResult.js`) which are quite similar

pyphilia:
Thanks for the review! I will apply the changes. 
I also thought it would be nice to have a generic function, but maybe we would like to do some processing or complete with more data in the case of result. In any case I think we will need two separate files even if we refactor a common processing function at some point. Let me know what you think.

hagoptaminian (approved):
perfect! let's keep `exportDataset.js` and `exportResult.js` separate for now

pyphilia:
After a notice from @Painguin I added the keys again, they are here for React to handle list of elements. I also added the missing translations.

src/langs/en.json
```diff
@@ -61,6 +61,8 @@
     "Information": "Information",
     "Back": "Back",
     "An unexpected error happened while opening the dataset.": "An unexpected error happened while opening the dataset.",
-    "Content": "Content"
+    "Content": "Content",
+    "export": "exporter",
```
Painguin, line 65: `"export": "export"`

src/langs/fr.json
```diff
@@ -61,6 +61,8 @@
     "Information": "Informations",
     "Back": "Retour",
     "An unexpected error happened while opening the dataset.": "Une erreur s'est produite lors de l'ouverture du dataset.",
-    "Content": "Contenu"
+    "Content": "Contenu",
+    "export": "export",
```
Painguin, line 65: `"export": "exporter"`

Painguin (changes requested):
Small mistake between french and english translation :laughing:

Painguin (approved):

Issue #45, opened 2020-10-15
Add View for Algorithm 


pyphilia:

Painguin:
We can view the algorithm through the edit algorithm page. I would say there is no need to add an additional view algorithm view/button.

hagoptaminian:
agree with @Painguin. I think this is now outdated.

PR #64, opened 2020-11-03
Add and edit algorithm

### LINKED ISSUE #46: Add Algorithm Button
This should be similar to the Dataset's Add button. 
A more advanced version of this button would be to include code editor to directly write the python code.

Painguin:
This PR implements the addition and edition of an algorithm.

When adding an algorithm, the user can either select a local file or write it in the app. Only python algorithms are supported for now.

The author of Graasp algorithms is `Graasp` and the author of user-created algorithms is `User`. Another possibility would have been to let the user specify the author name (let me know if you prefer it that way).

Editing a `Graasp` algorithm will create a copy of the file instead.

close #46


Painguin:
One feature we could also add is the possibility to run the algorithm on a dummy dataset while on the add or edit screen.

src/reducers/algorithmsReducer.js
```diff
@@ -5,20 +5,36 @@ import {
...
+      return state.updateIn(
+        ['current', 'activity'],
+        updateActivityList(payload),
+      );
     case GET_ALGORITHMS_SUCCESS:
       return state.setIn(['algorithms'], List(payload));
+    case GET_ALGORITHM_SUCCESS:
+      return state.setIn(['current', 'content'], Map(payload));
+    case GET_ALGORITHM_ERROR:
+      return state.setIn(['current', 'content'], List());
```
pyphilia, line 37: It should be `Map()`

src/components/Datasets.js
```diff
@@ -38,6 +38,10 @@ const styles = (theme) => ({
   infoAlert: {
     margin: theme.spacing(2),
   },
+  content: {
+    // adds bottom space so that button doesn't stay above table when fully scrolled
+    marginBottom: theme.spacing(10),
```
pyphilia, line 43: I approve this change, it improves the user experience. However, shouldn't it be added to the table element instead?

src/components/algorithm/AddAlgorithm.js
```diff
@@ -0,0 +1,234 @@
...
+import { AUTHOR_USER } from '../../config/constants';
+
+const styles = (theme) => ({
+  buttons: {
+    marginTop: theme.spacing(2),
+  },
+});
+
+const OPTION_FILE = 'OPTION_FILE';
+const OPTION_EDITOR = 'OPTION_EDITOR';
```
pyphilia, line 33: Maybe you can use something similar to
```
ADD_OPTIONS: {
  FILE: 'file',
  EDITOR: 'editor',
}
```
in `constants.js`

src/components/algorithm/AddAlgorithm.js
```diff
@@ -0,0 +1,234 @@
...
+
+  handleSave = () => {
+    const { dispatchAddAlgorithm } = this.props;
+    const { name, description, fileLocation, code, option } = this.state;
+    const author = AUTHOR_USER;
+    const payload =
+      option === OPTION_FILE
+        ? { name, description, author, fileLocation }
+        : { name, description, author, code };
+    if (name) {
```
pyphilia, line 98: Maybe instead of only dispatching if there is a name, have a form control and highlight the errors. https://material-ui.com/components/checkboxes/#CheckboxesGroup.js 
Or have something simple as disabling the save button if the mandatory fields are not filled in. (Or have both)

src/components/algorithm/AddAlgorithm.js
```diff
@@ -0,0 +1,234 @@
...
+    return (
+      <Main>
+        <Container>
+          <h1>{t('Add algorithm')}</h1>
+          <Grid container spacing={5}>
+            <Grid item xs={option === OPTION_FILE ? 5 : 7}>
+              <FormControl component="fieldset">
+                <RadioGroup
+                  aria-label="add option"
+                  name="add option"
```
pyphilia, line 116: translate

src/components/algorithm/AddAlgorithm.js
```diff
@@ -0,0 +1,234 @@
...
+                      options={{
+                        filters: [{ name: 'python', extensions: ['py'] }],
+                      }}
+                      onBrowseFileCallback={this.handleBrowseFileCallback}
+                    />
+                  </Grid>
+                </Grid>
+              ) : (
+                <Editor
+                  programmingLanguage="python"
```
pyphilia, line 156: Refactor `python`. It would be maybe nice to have `programmingLanguage` in the state to later use it to toggle language when we will support more of them.

src/components/algorithm/AddAlgorithm.js
```diff
@@ -0,0 +1,234 @@
...
+              </Button>
+            </Grid>
+          </Grid>
+        </Container>
+      </Main>
+    );
+  }
+}
+
+const mapStateToProps = () => ({});
```
pyphilia, line 217: Remove.

src/components/algorithm/EditAlgorithm.js
```diff
@@ -0,0 +1,248 @@
...
+        const payload = { name, description, author: AUTHOR_USER, code };
+        dispatchAddAlgorithm(payload);
+      } else {
+        const metadata = { ...algorithm?.toObject(), name, description };
+        dispatchSaveAlgorithm({ metadata, code });
+      }
+    }
+  };
+
+  renderBackButton = () => {
```
pyphilia, line 124: Maybe we should have a back button component if possible aha. What do you think?

src/components/algorithm/EditAlgorithm.js
```diff
@@ -0,0 +1,248 @@
...
+            <Alert severity="info" className={classes.infoAlert}>
+              {t(
+                `You are modifying a Graasp algorithm. Saving will create a new file instead.`,
+              )}
+            </Alert>
+          )}
+          <Grid container spacing={2} justify="center">
+            <Grid item xs={7}>
+              <Editor
+                programmingLanguage="python"
```
pyphilia, line 181: Refactor.

src/components/algorithm/EditAlgorithm.js
```diff
@@ -0,0 +1,248 @@
...
+          </Grid>
+        </Container>
+      </Main>
+    );
+  }
+}
+
+const mapStateToProps = ({ algorithms }) => ({
+  algorithm: algorithms.getIn(['current', 'content']),
+  activity: algorithms.getIn(['current', 'activity']).size > 0,
```
pyphilia, line 228: I like better  `Boolean(algorithms.getIn(['current', 'activity']).size)`

pyphilia (changes requested):
Nice work! Below some comments:

- Each time you click `save` in the add algorithm tab, it creates a new algorithm. While it is logic, I would prefer to "edit" the current one rather than creating a new algorithm (and the fields remain filled so it is a bit misleading). Maybe once created you can redirect on the edit tab, otherwise refresh all the fields to create a new algorithm. Maybe we can have additional buttons such as "Save and Continue editing" "Save and Quit" "Save and create new" (maybe not all of them but one-two more). The same behaviour happens when editing a Graasp Algorithm (multiple instances are created).

- Yes maybe submitting the author's name would be nice. Even though it might not really be used, but I think it is not too much work to add it anyway.

Just a thought: What if we have an internal python file with the recurrent functions (`load_dataset`, `save_dataset`,...) maybe this file would be available at some place in the interface as a "utils" file and the user would be able to edit it?

src/components/Datasets.js
Painguin: I don't think it should be in the table element. The margin is only there because of the add button and not all the pages have have it. The accommodated space amount might also be bigger or smaller if the page has another fixed component. So I think the component using the table is the one that should handle the space issues.

src/components/algorithm/EditAlgorithm.js
Painguin: Ah good call!

Painguin:
Thank you for your review @pyphilia ! 

The python utils file is a good idea! It will reduce the size of the python template and let the user focus on its algorithm. I'll add it as an issue.
I am still unsure about the use of the author field. I think each app will only be used by one person and the main use of the author I see is to differentiate its own algorithms from the offered ones. But you are right adding it would be very little work.

pyphilia:
I am not sure either for its utility, but it might be interesting when you use algorithm you don't really own? We can let it at it is or discuss later.

src/actions/algorithms.js
```diff
@@ -89,13 +89,16 @@ export const saveAlgorithm = (algorithm) => (dispatch) => {
...
-export const addAlgorithm = (algorithm) => (dispatch) => {
+export const addAlgorithm = ({ payload, onSuccess }) => (dispatch) => {
   const flagAddingAlgorithm = createFlag(FLAG_ADDING_ALGORITHM);
   try {
     dispatch(flagAddingAlgorithm(true));
-    window.ipcRenderer.send(ADD_ALGORITHM_CHANNEL, algorithm);
+    window.ipcRenderer.send(ADD_ALGORITHM_CHANNEL, payload);
     window.ipcRenderer.once(ADD_ALGORITHM_CHANNEL, async (event, response) => {
       dispatch(response);
+      if (onSuccess) {
```
pyphilia, line 99: You can use `onSuccess?.()`

src/components/common/BackButton.js
```diff
@@ -0,0 +1,42 @@
...
+  className: PropTypes.string,
+};
+
+BackButton.defaultProps = {
+  className: null,
+};
+
+const TranslatedComponent = withTranslation()(BackButton);
+
+export default withRouter(TranslatedComponent);
```
pyphilia, line 42: 👍

pyphilia (approved):

Painguin:
In the same way how the dataset might not need both a view and edit button like as you discussed, the algorithm might also not need both of them. So maybe we could close #45 as a result.

src/components/algorithm/EditAlgorithm.js
```diff
@@ -0,0 +1,232 @@
...
+    if (algorithm && algorithm !== prevAlgorithm) {
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({
+        name: algorithm.get('name'),
+        description: algorithm.get('description'),
+        code: algorithm.get('code'),
+      });
+    }
+  }
+
```
pyphilia, line 87: Add `componentWillUnMount` to clear the state (remove the current algorithm data) to go along with the issue #66

pyphilia (changes requested):

pyphilia (approved):

hagoptaminian (changes requested):
@Painguin very nice and works well... two small comments:

1. Make the new page's heading capitalized ('Add **A**lgorithm')
2. I think the Back button would look better in the bottom right of the screen (where the Plus icon for adding an algorithm is). That way similar functionality is displayed consistently, _and_ the new elements that appear on the page are all dedicated to the function the user asked for (adding an algorithm)

PR #68, opened 2020-11-04
test: add dataset test

### LINKED ISSUE #63: Test Dataset tab behaviour

pyphilia:
These tests are very basic. They test only the very basic feature and very few cases. Since this application is a prototype, I propose to let them as it is, unless you have precise cases you want to cover. Ideally each time a bug is found we should fix it and have a dedicated test to confirm the fix.

close #63


Painguin (approved):
Looking good :+1:

Painguin (approved):

PR #72, opened 2020-11-05
feat: add name field for execution form

### LINKED ISSUE #70: Add name field in execution form

pyphilia:
close #70


pyphilia:
I found one bug, I will update soon. This PR will also include the first part of #71

pyphilia:
I updated the PR. This PR adds the name for the execution form and the dataset select now also proposes results.
Screenshot 2020-11-06 at 10 43 00
Screenshot 2020-11-06 at 10 43 21

I did not went for a grouping by original dataset as it requires changes I already did in the execution table PR. If you don't mind I propose this simple solution for the MVP.

src/components/Executions.js
```diff
@@ -71,149 +34,16 @@ class Executions extends Component {
...
-            >
-              {algorithms.map(({ id, name }) => (
-                <MenuItem value={id} key={id}>
-                  {name}
-                </MenuItem>
-              ))}
-            </Select>
-          </FormControl>
-          {this.renderExecuteButton()}
+          <AddExecutionForm />
```
hagoptaminian, line 46: nicely done to separate this into another component

src/components/execution/AddExecutionForm.js
```diff
@@ -0,0 +1,288 @@
...
+      .sortBy(({ name }) => name)
+      .map(({ id, name }) => (
+        <MenuItem value={id} key={id} className={classes.menuItem}>
+          {name}
+        </MenuItem>
+      ));
+
+    return (
+      <FormControl variant="outlined" className={classes.formControl}>
+        <InputLabel id="dataset-select">
```
hagoptaminian, line 142: so the `<InputLabel>`s have this ugly behavior:

Screen Shot 2020-11-09 at 11 17 40 AM

a quick fix would be to add a `backgroundColor: '#F7F7F7'` to them (same as the background color of `BrowserWindow`):

Screen Shot 2020-11-09 at 11 18 33 AM

public/app/listeners/executePythonAlgorithm.js
```diff
@@ -35,14 +35,13 @@ const createNewResultDataset = ({
 
 const executePythonAlgorithm = (mainWindow, db) => (
   event,
-  { datasetId, algorithmId },
+  { sourceId, algorithmId, filename },
```
hagoptaminian, line 38: maybe worth it here and in `AddExecutionForm.js` and `/actions/executions.js` to use something like `userProvidedFilename` instead of `filename`

hagoptaminian (changes requested):
nice one @pyphilia, left some small comments

src/components/execution/AddExecutionForm.js
pyphilia: I agree, I will do as you mentioned.

src/components/execution/AddExecutionForm.js
```diff
@@ -0,0 +1,288 @@
...
+    if (language) {
+      dispatchExecuteAlgorithm({ sourceId, algorithmId, language, filename });
+    }
+  };
+
+  renderDatasetsAndResultsSelect = () => {
+    const { sourceId } = this.state;
+    const { classes, t, datasets, results } = this.props;
+
+    if (datasets?.isEmpty() || results?.isEmpty()) {
```
Painguin, line 120: should be `&&`

src/components/execution/AddExecutionForm.js
```diff
@@ -0,0 +1,288 @@
...
+    return (
+      <FormControl variant="outlined" className={classes.formControl}>
+        <InputLabel id="dataset-select">
+          {`${t('Dataset')} ${t('(Required)')}`}
+        </InputLabel>
+        <Select
+          labelId="dataset-select"
+          value={sourceId}
+          onChange={this.handleDatasetSelectOnChange}
+          label={t('Dataset')}
```
Painguin, line 149: The issue @hagoptaminian mentioned where `(required)` is crossed out is caused by this `label` prop value not corresponding to the one in the `InputLabel`. Replace `t('Dataset')` by `` `${t('Dataset')} ${t('(Required)')}` ``.

src/components/execution/AddExecutionForm.js
```diff
@@ -0,0 +1,288 @@
...
+        {this.renderDatasetsAndResultsSelect()}
+        <FormControl variant="outlined" className={classes.formControl}>
+          <InputLabel id="algorithm-select">
+            {`${t('Algorithm')} ${t('(Required)')}`}
+          </InputLabel>
+          <Select
+            labelId="algorithm-select"
+            value={algorithmId}
+            onChange={this.handleAlgorithmSelectOnChange}
+            label={t('Algorithm')}
```
Painguin, line 236: Replace `t('Algorithm')` by `` `${t('Algorithm')} ${t('(Required)')}` ``

Painguin (changes requested):
Nice work @pyphilia. Left small comments.

src/components/execution/AddExecutionForm.js
Painguin: As mentioned in my last review, This is caused by the `label` prop value in the select below not matching the one of the `InputLabel`

Issue #66, opened 2020-11-04
Clear state when exiting single algorithm, dataset, result views


pyphilia:
If the state is not cleared before exiting a view page, the next time the page is accessed with a different id, it will display the previous one before.

hagoptaminian:
also make sure state is cleared in visualizations (https://github.com/graasp/graasp-insights/pull/74#discussion_r522100983)

PR #74, opened 2020-11-10
feat: add standard visualizations

hagoptaminian:
these are the same charts found in graasp-research, and look as follows in -insights:

Screen Shot 2020-11-10 at 3 32 40 PM

a few notes:

1. instead of `mapDispatchToprops` and `mapStateToProps` I used react-redux hooks in the new `Visualizations.js` component
2. you will need to add a `REACT_APP_GOOGLE_KEY=` to `.env.local` for the map to show up without the 'developer' overlay; I can share our key offline
3. a lot of this is adapted/reused from graasp-research (so feedback counts double!)

### DIFF
--- a/src/components/Visualizations.js
+++ b/src/components/Visualizations.js
@@ -1,0 +1,67 @@
+import React, { useState, useEffect } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import { makeStyles } from '@material-ui/core/styles';
+import { useTranslation } from 'react-i18next';
+import Alert from '@material-ui/lab/Alert';
+import Main from './common/Main';
+import AddVisualizationForm from './charts/AddVisualizationForm';
+import ChartsLayout from './charts/ChartsLayout';
+import { getDatasets, getDataset } from '../actions';
+
+const useStyles = makeStyles((theme) => ({
+  infoAlert: {
+    margin: theme.spacing(2),
+  },
+}));
+
+const Visualizations = () => {
+  // i18next, material-ui, and react-redux hooks
+  const { t } = useTranslation();
+  const classes = useStyles();
+  const dispatch = useDispatch();
+
+  // initialize component state holding the selected dataset to be visualized
+  const [selectedDatasetId, setSelectedDatasetId] = useState('');
+
+  // on component load, dispatch getDatasets action
+  useEffect(() => {
+    dispatch(getDatasets);
+  }, [dispatch]);
+
+  // retrieve datasets from redux store
+  // (note: react-redux useSelector roughly equivalent to mapStateToProps)
+  const datasets = useSelector((state) => state.dataset.get('datasets'));
+  // retrieve selected dataset's content from redux store
+  const retrievedDatasetContent = useSelector((state) =>
+    state.dataset.getIn(['current', 'content']).get('content'),
+  );
+
+  // function triggered when 'Visualize' button is clicked
+  const fetchDatasetToBeVisualized = () => {
+    dispatch(getDataset({ id: selectedDatasetId }));
+  };
+
+  if (!datasets.size) {
+    return (
+      <Main>
+        <Alert severity="info" className={classes.infoAlert}>
+          {t('Load a dataset first!')}
+        </Alert>
+      </Main>
+    );
+  }
+
+  return (
+    <Main>
+      <AddVisualizationForm
+        selectedDatasetId={selectedDatasetId}
+        setSelectedDatasetId={setSelectedDatasetId}
+        fetchDatasetToBeVisualized={fetchDatasetToBeVisualized}
+        datasets={datasets}
+      />
+      <ChartsLayout retrievedDatasetContent={retrievedDatasetContent} />
+    </Main>
+  );
+};
+
+export default Visualizations;
--- a/src/components/Visualizations.js
+++ b/src/components/Visualizations.js
@@ -25,7 +25,7 @@ const Visualizations = () => {
 
   // on component load, dispatch getDatasets action
   useEffect(() => {
-    dispatch(getDatasets);
+    dispatch(getDatasets());
   }, [dispatch]);
 
   // retrieve datasets from redux store
--- a/src/components/charts/ActionsByDayChart.js
+++ b/src/components/charts/ActionsByDayChart.js
@@ -19,77 +19,88 @@ import {
   filterActionsByUser,
   findYAxisMax,
 } from '../../utils/charts';
-import { CONTAINER_HEIGHT } from '../../config/constants';
+import { TICK_FONT_SIZE, CONTAINER_HEIGHT } from '../../config/constants';
 
-const useStyles = makeStyles(() => ({
-  typography: { textAlign: 'center' },
+const useStyles = makeStyles((theme) => ({
+  typography: { textAlign: 'center', marginTop: theme.spacing(4) },
+  chart: { marginTop: 30, marginRight: 20, marginBottom: 20, marginLeft: 20 },
 }));
 
 const ActionsByDayChart = ({
   actions,
   allUsersConsolidated,
   usersToFilter,
+  setError,
 }) => {
   const { t } = useTranslation();
   const theme = useTheme();
   const classes = useStyles();
 
-  // actionsByDay is the object passed, after formatting, to the BarChart component below
-  // (1) if you remove all names in the react-select dropdown, usersToFilter becomes null; in this case, show all actions
-  // (2) if no users are selected (i.e. usersToFilter.length === 0), show all actions
-  // (3) if all users are selected (i.e. usersToFilter.length === allUsers.length), also show all actions
-  // #3 above is necessary: some actions are made by users NOT in the users list (e.g. user account deleted)
-  // e.g. we retrieve 100 total actions and 10 users, but these 10 users have only made 90 actions
-  // therefore, to avoid confusion: when all users are selected, show all actions
-  let actionsByDay;
-  if (
-    usersToFilter === null ||
-    usersToFilter.length === 0 ||
-    usersToFilter.length === allUsersConsolidated.length
-  ) {
-    actionsByDay = getActionsByDay(actions);
-  } else {
-    actionsByDay = getActionsByDay(filterActionsByUser(actions, usersToFilter));
-  }
+  try {
+    // actionsByDay is the object passed, after formatting, to the BarChart component below
+    // (1) if you remove all names in the react-select dropdown, usersToFilter becomes null; in this case, show all actions
+    // (2) if no users are selected (i.e. usersToFilter.length === 0), show all actions
+    // (3) if all users are selected (i.e. usersToFilter.length === allUsers.length), also show all actions
+    // #3 above is necessary: some actions are made by users NOT in the users list (e.g. user account deleted)
+    // e.g. we retrieve 100 total actions and 10 users, but these 10 users have only made 90 actions
+    // therefore, to avoid confusion: when all users are selected, show all actions
+    let actionsByDay;
+    if (
+      usersToFilter === null ||
+      usersToFilter.length === 0 ||
+      usersToFilter.length === allUsersConsolidated.length
+    ) {
+      actionsByDay = getActionsByDay(actions);
+    } else {
+      actionsByDay = getActionsByDay(
+        filterActionsByUser(actions, usersToFilter),
+      );
+    }
 
-  const yAxisMax = findYAxisMax(actionsByDay);
-  const formattedActionsByDay = formatActionsByDay(actionsByDay);
+    const yAxisMax = findYAxisMax(actionsByDay);
+    const formattedActionsByDay = formatActionsByDay(actionsByDay);
 
-  // if selected user(s) have no actions, render component with message that there are no actions
-  if (formattedActionsByDay.length === 0) {
+    // if selected user(s) have no actions, render component with message that there are no actions
+    if (formattedActionsByDay.length === 0) {
+      return (
+        <EmptyChart
+          usersToFilter={usersToFilter}
+          chartTitle={t('Actions by Day')}
+        />
+      );
+    }
+
     return (
-      <EmptyChart
-        usersToFilter={usersToFilter}
-        chartTitle={t('Actions by Day')}
-      />
+      <>
+        <Typography variant="subtitle1" className={classes.typography}>
+          {t('Actions by Day')}
+        </Typography>
+        <ResponsiveContainer width="95%" height={CONTAINER_HEIGHT}>
+          <BarChart data={formattedActionsByDay} className={classes.chart}>
+            <CartesianGrid strokeDasharray="2" />
+            <XAxis dataKey="date" tick={{ fontSize: TICK_FONT_SIZE }} />
+            <YAxis tick={{ fontSize: TICK_FONT_SIZE }} domain={[0, yAxisMax]} />
+            <Tooltip />
+            <Bar
+              dataKey="count"
+              name="Count"
+              fill={theme.palette.primary.main}
+            />
+          </BarChart>
+        </ResponsiveContainer>
+      </>
     );
+  } catch (err) {
+    setError(err);
+    return null;
   }
-
-  return (
-    <>
-      <Typography variant="subtitle1" className={classes.typography}>
-        {t('Actions by Day')}
-      </Typography>
-      <ResponsiveContainer width="95%" height={CONTAINER_HEIGHT}>
-        <BarChart
-          data={formattedActionsByDay}
-          margin={{ top: 30, bottom: 20, left: 20, right: 20 }}
-        >
-          <CartesianGrid strokeDasharray="2" />
-          <XAxis dataKey="date" tick={{ fontSize: 14 }} />
-          <YAxis tick={{ fontSize: 14 }} domain={[0, yAxisMax]} />
-          <Tooltip />
-          <Bar dataKey="count" name="Count" fill={theme.palette.primary.main} />
-        </BarChart>
-      </ResponsiveContainer>
-    </>
-  );
 };
 
 ActionsByDayChart.propTypes = {
   actions: PropTypes.arrayOf(PropTypes.object).isRequired,
   allUsersConsolidated: PropTypes.arrayOf(PropTypes.object).isRequired,
   usersToFilter: PropTypes.arrayOf(PropTypes.object).isRequired,
+  setError: PropTypes.func.isRequired,
 };
 
 export default ActionsByDayChart;
--- a/src/components/charts/ActionsByDayChart.js
+++ b/src/components/charts/ActionsByDayChart.js
@@ -1,0 +1,95 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useTranslation } from 'react-i18next';
+import { makeStyles, useTheme } from '@material-ui/core/styles';
+import Typography from '@material-ui/core/Typography';
+import {
+  BarChart,
+  CartesianGrid,
+  XAxis,
+  YAxis,
+  Tooltip,
+  Bar,
+  ResponsiveContainer,
+} from 'recharts';
+import EmptyChart from './EmptyChart';
+import {
+  getActionsByDay,
+  formatActionsByDay,
+  filterActionsByUser,
+  findYAxisMax,
+} from '../../utils/charts';
+import { CONTAINER_HEIGHT } from '../../config/constants';
+
+const useStyles = makeStyles(() => ({
+  typography: { textAlign: 'center' },
+}));
+
+const ActionsByDayChart = ({
+  actions,
+  allUsersConsolidated,
+  usersToFilter,
+}) => {
+  const { t } = useTranslation();
+  const theme = useTheme();
+  const classes = useStyles();
+
+  // actionsByDay is the object passed, after formatting, to the BarChart component below
+  // (1) if you remove all names in the react-select dropdown, usersToFilter becomes null; in this case, show all actions
+  // (2) if no users are selected (i.e. usersToFilter.length === 0), show all actions
+  // (3) if all users are selected (i.e. usersToFilter.length === allUsers.length), also show all actions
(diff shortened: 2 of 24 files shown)


src/components/charts/ActionsByDayChart.js
```diff
@@ -0,0 +1,95 @@
...
+      <Typography variant="subtitle1" className={classes.typography}>
+        {t('Actions by Day')}
+      </Typography>
+      <ResponsiveContainer width="95%" height={CONTAINER_HEIGHT}>
+        <BarChart
+          data={formattedActionsByDay}
+          margin={{ top: 30, bottom: 20, left: 20, right: 20 }}
+        >
+          <CartesianGrid strokeDasharray="2" />
+          <XAxis dataKey="date" tick={{ fontSize: 14 }} />
```
pyphilia, line 79: Refactor fontSize, in `useStyles` and have TICK_FONT_SIZE in constant? I think there is a constant in `theme` as well but I am not sure.

src/components/charts/ActionsByDayChart.js
```diff
@@ -0,0 +1,95 @@
...
+          <YAxis tick={{ fontSize: 14 }} domain={[0, yAxisMax]} />
```
pyphilia, line 80: Same as above.

src/components/charts/ActionsByDayChart.js
```diff
@@ -0,0 +1,95 @@
...
+
+  return (
+    <>
+      <Typography variant="subtitle1" className={classes.typography}>
+        {t('Actions by Day')}
+      </Typography>
+      <ResponsiveContainer width="95%" height={CONTAINER_HEIGHT}>
+        <BarChart
+          data={formattedActionsByDay}
+          margin={{ top: 30, bottom: 20, left: 20, right: 20 }}
```
pyphilia, line 76: Would it be worth to refactor it?

src/components/charts/ActionsByTimeOfDayChart.js
```diff
@@ -0,0 +1,101 @@
...
+
+  return (
+    <>
+      <Typography variant="subtitle1" className={classes.typography}>
+        {t('Actions by Time of Day')}
+      </Typography>
+      <ResponsiveContainer width="95%" height={CONTAINER_HEIGHT}>
+        <BarChart
+          data={formattedActionsByTimeOfDay}
+          margin={{ top: 30, bottom: 20, left: 20, right: 20 }}
```
pyphilia, line 82: Same comments as for `ActionsByDayChart`

src/components/charts/ActionsByTimeOfDayChart.js
```diff
@@ -0,0 +1,101 @@
...
+        >
+          <CartesianGrid strokeDasharray="2" />
+          <XAxis dataKey="timeOfDay" tick={{ fontSize: 14 }} />
+          <YAxis tick={{ fontSize: 14 }} domain={[0, yAxisMax]} />
+          <Tooltip content={<CustomTooltip />} />
+          <Bar dataKey="count" name="Count" fill="#8884d8" />
```
pyphilia, line 88: Use theme.

src/components/charts/ActionsByVerbChart.js
```diff
@@ -0,0 +1,90 @@
...
+      <Typography variant="subtitle1" className={classes.typography}>
+        {t('Actions by Verb')}
+      </Typography>
+      <ResponsiveContainer width="95%" height={CONTAINER_HEIGHT}>
+        <PieChart fontSize={14}>
+          <Pie
+            data={formattedActionsByVerb}
+            dataKey="percentage"
+            nameKey="verb"
+            fill="#82ca9d"
```
pyphilia, line 69: Refactor?

src/components/charts/ActionsMap.js
```diff
@@ -0,0 +1,180 @@
...
+  CLUSTER_RADIUS,
+  ENTER_KEY_CODE,
+} from '../../config/constants';
+
+const useStyles = makeStyles((theme) => ({
+  clusterMarker: {
+    color: '#fff',
+    background: theme.palette.primary.main,
+    borderRadius: '50%',
+    padding: 12,
```
pyphilia, line 24: Use `theme.spacing`

src/components/charts/ActionsMap.js
```diff
@@ -0,0 +1,180 @@
...
+              cluster: isCluster,
+              point_count: pointCount,
+            } = cluster.properties;
+            if (isCluster) {
+              return (
+                <Marker key={cluster.id} lat={latitude} lng={longitude}>
+                  <div
+                    className={classes.clusterMarker}
+                    style={{
+                      width: `${calculateClusterRadius(
```
pyphilia, line 139: Might be worth defining this value above in the code since it is used twice.

src/components/charts/UsersSelect.js
```diff
@@ -0,0 +1,93 @@
...
+    <div className={classes.container}>
+      <Select
+        styles={customStyles}
+        options={[allOption, ...allUsersConsolidated]}
+        isMulti
+        closeMenuOnSelect={false}
+        hideSelectedOptions={false}
+        allowSelectAll
+        getOptionLabel={(option) => option.name}
+        placeholder="Filter by user..."
```
pyphilia, line 68: translate.

src/components/charts/custom/CustomValueContainer.js
```diff
@@ -0,0 +1,44 @@
...
+    const alsoSelectedCount = length - 2;
+    if (length === 1) {
+      values = `${values[0].key}`;
+    } else if (length === 2) {
+      values = `${values[0].key}, ${values[1].key}`;
+    } else {
+      const firstSelection = values[0].key;
+      const secondSelection = values[1].key;
+      values = t('Many users selected...', {
+        firstSelection,
```
pyphilia, line 27: I didn't know `t` could take parameters! What do `firstSelection` and `secondSelection`?

src/utils/charts.js
```diff
@@ -0,0 +1,241 @@
...
+    [LATE_NIGHT]: 0,
+    [EARLY_MORNING]: 0,
+    [MORNING]: 0,
+    [AFTERNOON]: 0,
+    [EVENING]: 0,
+    [NIGHT]: 0,
+  };
+  actions.forEach((action) => {
+    const actionHourOfDay = getActionHourOfDay(action);
+    if (actionHourOfDay >= 0 && actionHourOfDay < 4) {
```
pyphilia, line 82: Minor comment: maybe defining the limits for each period in constants would help in case we would like to have less categories, etc..

pyphilia (changes requested):
Looks very nice! 

Some thoughts:
- It might require a lot of work, but what do you think of having a quick action in Dataset which would redirect to the corresponding page? Instead of a dedicated tab. But a dedicated page can do the job as well.
- Until we have schema validation (to identify Graasp structured datasets), you will have to error proof your code. For instance a JSON file containing `{ "name": "hello" }` should be display some alert ("No visualisation is available for this dataset") (Keep in mind that viewing this file in the interface currently throws error but is resolved in the edit dataset PR). Maybe try to take a "normal" dataset and remove fields bit by bit to detect different cases.
- It might be worth to update the `README.md` file with the necessary env variables.
- On first load (if you reload directly from Vizualisations) it does not fetch the datasets. I can see the alert `Load a dataset first!` even though I have datasets saved. Once I navigate to another tab and come back the dataset select is displayed.

src/components/charts/custom/CustomValueContainer.js
hagoptaminian: I had forgotten to update `en.json` and `fr.json` with the translation pairs (fixed this), but basically what this does is allow you to pass variables into your translations. so in `en.json` I add:

`"Many users selected...": "{{firstSelection}}, {{secondSelection}}, and {{alsoSelectedCount}} other{{plural}} selected"`

--> If there is an error with the variables or the translation pair is missing, it will default to _'Many users selected...'_, otherwise it will display: _'X, Y, and N others selected'_

src/components/charts/ActionsByDayChart.js
hagoptaminian: good call, factored out TICK_FONT_SIZE into a constant. as for using useStyles/material-ui: this has to explicitly be a `tick={{...}}` prop on the recharts `XAxis` and `YAxis` components (rather than an added class). so `className={classes.tick}` would not work

src/components/charts/ActionsByDayChart.js
hagoptaminian: any suggestions? you mean to factor out the margins as constants?

src/components/charts/ActionsByVerbChart.js
hagoptaminian: good catch, not sure what this `fill` is doing here. I removed it. it doesn't have an effect

src/utils/charts.js
hagoptaminian: good idea, but I think tricky in practice. if we have something like `START_OF_DAY = 0`, `END_OF_MORNING = 4`, etc, that might work, but then both this `if/else` chain and the `LATE_NIGHT`, `EARLY_MORNING`, etc, constants would have to be refactored/redefined, so it gets a bit circular.

src/components/charts/ActionsByDayChart.js
pyphilia: I would simply use `classes` ^^

hagoptaminian: 👍

hagoptaminian:
@pyphilia thanks for the comments:

1. I added some basic error handling which will catch a lot of invalid datasets but as discussed this doesn't cover every single possibility (need to sync with @Painguin's work on schema validation)

2. I'm open to the idea of accessing the visualizations from a quick action, but for now left it as is (as a tab). let's discuss if/how to modify later

src/utils/charts.js
```diff
@@ -0,0 +1,241 @@
...
+    ]);
+  }
+
+  // convert to recharts required format
+  return formattedActionsByVerbArray.map((entry) => {
+    return {
+      verb: entry[0],
+      percentage: entry[1],
+    };
+  });
```
Painguin, line 164:
``` 
return formattedActionsByVerbArray.map(([verb, percentage]) => {
  return {
    verb,
    percentage,
  };
});
```

src/components/Visualizations.js
```diff
@@ -0,0 +1,67 @@
...
+        setSelectedDatasetId={setSelectedDatasetId}
+        fetchDatasetToBeVisualized={fetchDatasetToBeVisualized}
+        datasets={datasets}
+      />
+      <ChartsLayout retrievedDatasetContent={retrievedDatasetContent} />
+    </Main>
+  );
+};
+
+export default Visualizations;
```
Painguin, line 67: In line with #66 the state of the dataset content needs to be cleared when the component is unmounted

Painguin (changes requested):
The visualizations look very good @hagoptaminian :open_mouth:. Nice work! Here are some comments.

src/langs/fr.json
```diff
@@ -63,6 +63,18 @@
...
+    "No actions to show for these users": "",
+    "No actions to show for this user": "",
+    "Visualize": "Visualiser",
+    "Actions by Location": "",
+    "Actions by Verb": "",
+    "Actions by Time of Day": "",
+    "Actions by Day": "",
+    "Many users selected...": "{{firstSelection}}, {{secondSelection}}, et {{alsoSelectedCount}} autre{{plural}}...",
+    "There was an error visualizing this dataset.": "",
+    "This dataset does not contain any actions.": ""
```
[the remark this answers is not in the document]
hagoptaminian: merci!

src/components/charts/AddVisualizationForm.js
```diff
@@ -0,0 +1,91 @@
...
+    margin: theme.spacing(1),
+    width: '100%',
+  },
+  select: {
+    padding: 0,
+    width: '300px',
+  },
+  inputLabel: {
+    backgroundColor: '#F7F7F7',
+  },
```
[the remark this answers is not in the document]
hagoptaminian: you're right, in this case it doesn't make a difference, because the text in the `<InputLabel>` is short. but see here why it does make a difference: https://github.com/graasp/graasp-insights/pull/72#discussion_r519698356. so I'll keep it just in case this text becomes say _'Dataset (Required)'_

src/utils/charts.js
```diff
@@ -0,0 +1,241 @@
...
+  return actionsByDay;
+};
+
+// Takes object with {key: value} pairs of {date: #-of-actions} and returns a date-sorted array in Recharts.js format
+export const formatActionsByDay = (actionsByDayObject) => {
+  const actionsByDayArray = Object.entries(actionsByDayObject);
+  const sortedActionsByDay = actionsByDayArray.sort(
+    (entryA, entryB) => Date.parse(entryA[0]) - Date.parse(entryB[0]),
+  );
+  return sortedActionsByDay.map((entry) => {
```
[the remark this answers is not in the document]
hagoptaminian: 👍 👍

src/utils/charts.js
```diff
@@ -0,0 +1,241 @@
...
+  });
+};
+
+// given an actionsByDay object, finds max value to set on the yAxis in the graph in ActionsByDayChart.js
+export const findYAxisMax = (actionsByDay) => {
+  const arrayOfActionsCount = Object.values(actionsByDay);
+  if (!arrayOfActionsCount.length) {
+    return null;
+  }
+  const maxActionsCount = arrayOfActionsCount.reduce((a, b) => Math.max(a, b));
```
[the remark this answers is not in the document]
hagoptaminian: good idea! this led me to https://medium.com/coding-at-dawn/the-fastest-way-to-find-minimum-and-maximum-values-in-an-array-in-javascript-2511115f8621... so will keep as is given better `.reduce` performance vs. `Math.max`

src/utils/charts.js
```diff
@@ -0,0 +1,241 @@
...
+// Takes object with {key: value} pairs of {timeOfDay: #-of-actions}
+// returns a date-sorted array in Recharts.js format
+export const formatActionsByTimeOfDay = (actionsByTimeOfDayObject) => {
+  const actionsByTimeOfDayArray = Object.entries(actionsByTimeOfDayObject);
+  return actionsByTimeOfDayArray.map((entry) => {
+    return {
+      timeOfDay: entry[0],
+      count: entry[1],
+    };
+  });
```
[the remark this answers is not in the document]
hagoptaminian: 👍

src/components/Visualizations.js
hagoptaminian: yes, good call. looks like @pyphilia has added some functionality to do this in #75. let's resolve this issue alongside the rest in #66

src/utils/charts.js
```diff
@@ -0,0 +1,241 @@
...
+  actions.forEach((action) => {
+    if (!actionsByVerb[action.verb]) {
+      // if verb is still not in the actionsByVerb object, add it and assign it to (1 / totalActions)
+      // we use (1 / totalActions) because in the end we want this object to be {verb: PERCENTAGE-of-total-actions}
+      actionsByVerb[action.verb] = 1 / totalActions;
+    } else {
+      actionsByVerb[action.verb] += 1 / totalActions;
+    }
+  });
+  return actionsByVerb;
```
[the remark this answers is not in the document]
hagoptaminian: interesting... could do this. but downside is that we'd have to do `Object.entries()` on `actionsByVerb`, loop through that array to divide each verb's actionCount by totalActions count, and then re-convert into an object... what do you think?

src/components/charts/custom/CustomValueContainer.js
```diff
@@ -0,0 +1,44 @@
...
+// eslint-disable-next-line react/prop-types
+const CustomValueContainer = ({ children, ...props }) => {
+  const { t } = useTranslation();
+
+  // values is an array of objects corresponding to the currently made selection
+  let [values] = children;
+
+  if (Array.isArray(values)) {
+    const { length } = values;
+    const plural = length === 3 ? '' : 's';
```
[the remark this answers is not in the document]
hagoptaminian: such a good point! the translation files are just json files, so we can't pass booleans into them. there are two possible solutions:

- we can detect active language in this component via `import i18n from 'i18next'` then `i18n.language`, and then use a switch statement on language which assigns the plural variable depending on the language
- simpler: the multiple selection translation becomes: `A, B, and N other(s) selected...` (and similar in other languages), which would cover both N=1 and N > 1. that way we don't even pass a `plural` parameter

for now I'll leave as is and create an issue to investigate best approach further

hagoptaminian:
@Painguin thanks for the feedback! very helpful. let me know what you think of the two comments above

src/components/charts/AddVisualizationForm.js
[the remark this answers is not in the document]
Painguin: Oh that was caused by the label prop of the select being only `Dataset` while the `InputLabel` value was `Dataset (required)`. So if we match both of them to `Dataset (required)` there shouldn't be any issues. This is some weird behavior but I guess they just check the length of the label prop for the hiding parts of the bounding box and display the value of the `InputLabel`

src/components/charts/custom/CustomValueContainer.js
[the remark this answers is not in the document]
Painguin: https://www.i18next.com/translation-function/plurals

It seems we can append `_plural` to the key and pass a `count` value in the `t(..., {..., count: n})` call.
The following seems to work:

In CustomValueContainer:
```js
values = t('Many users selected...', {
  firstSelection,
  secondSelection,
  count: alsoSelectedCount,
});
```

In en.json:
```json
"Many users selected...": "{{firstSelection}}, {{secondSelection}}, and one other selected",
"Many users selected..._plural": "{{firstSelection}}, {{secondSelection}}, and {{count}} others selected",
```

In fr.json:
```json
"Many users selected...": "{{firstSelection}}, {{secondSelection}}, et un autre...",
"Many users selected..._plural": "{{firstSelection}}, {{secondSelection}}, et {{count}} autres...",
```

src/utils/charts.js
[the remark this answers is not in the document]
Painguin: https://stackoverflow.com/questions/14810506/map-function-for-objects-instead-of-arrays
We could also iterate over the keys of the object and divide `actionsByVerb[verb] /= totalActions`. I does require a second (smaller, since we iterate over each distinct verb) loop but I think it is more correct.

src/utils/charts.js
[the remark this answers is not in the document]
Painguin: Oh I see... Good investigation!

src/components/charts/AddVisualizationForm.js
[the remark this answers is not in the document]
hagoptaminian: ah nice catch. did not see your response to that comment. you're right, fixed it

src/utils/charts.js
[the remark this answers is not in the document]
hagoptaminian: the `Object.fromEntries()` and `Object.entries()` pair work well together. fixed this.

src/components/charts/custom/CustomValueContainer.js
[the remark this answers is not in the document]
hagoptaminian: very cool! thanks for figuring it out

hagoptaminian:
thanks for the reviews and suggestions @Painguin @pyphilia . let me know when ok to rebase/merge

src/langs/fr.json
```diff
@@ -63,6 +63,19 @@
...
+    "No actions to show for these users": "Aucune action à afficher pour ces utilisateurs",
+    "No actions to show for this user": "Aucune action à afficher pour cet utilisateur",
+    "Visualize": "Visualiser",
+    "Actions by Location": "Actions par localisation",
+    "Actions by Verb": "Actions par verbe",
+    "Actions by Time of Day": "Actions par moment de la journée",
+    "Actions by Day": "Actions par jour",
+    "Many users selected...": "{{firstSelection}}, {{secondSelection}}, et un autre...",
+    "Many users selected..._plural": "{{firstSelection}}, {{secondSelection}}, et {{count}} autres...",
+    "There was an error visualizing this dataset.": "Une erreur s'est produise en visualisant ce dataset",
```
pyphilia, line 78: `produise` to `produite` (t not s)

src/utils/charts.js
```diff
@@ -0,0 +1,255 @@
...
+      count,
+    };
+  });
+};
+
+// Takes array of action objects and returns an object with {key: value} pairs of {verb: %-of-actions}
+export const getActionsByVerb = (actions) => {
+  const totalActionsCount = actions.length;
+  const actionsByVerbCount = {};
+  actions.forEach((action) => {
```
pyphilia, line 118: You don't have to update this, but this could have been done using either:
- groupby (lodash) on the actions verb would give you a map whose keys would be the actions and the values would be an array of all corresponding actions. Then a map can be used to have the length of the array.
- reduce: given an accumulator `{ }` you can go through `actions` and add them to the map "one by one". It looks like a lot to your foreach procedure but the accumulator is a bit more transparent.

I prefer the first option as it is much more clearer.

src/utils/charts.js
```diff
@@ -0,0 +1,255 @@
...
+            (acc, current) => acc + current[1],
+            0,
+          )
+        ).toFixed(2),
+      ),
+    ]);
+  }
+
+  // convert to recharts required format
+  return formattedActionsByVerbArray.map(([verb, percentage]) => {
```
pyphilia, line 173: Just a remark as well: you can refactor it as `([verb, percentage]) => ({verb, percentage})`. The parenthesis allow to return object without errors.

pyphilia (approved):
Looks good to me! I left you some comments about refactor you can or cannot take it into account and merge.

src/utils/charts.js
hagoptaminian: both interesting!

src/utils/charts.js
hagoptaminian: good call!

PR #75, opened 2020-11-11
fix: clear dataset and result state

### LINKED ISSUE #66: Clear state when exiting single algorithm, dataset, result views

pyphilia:
close #66


src/reducers/resultReducer.js
```diff
@@ -8,6 +8,7 @@ import {
   FLAG_EXPORTING_RESULT,
   FLAG_DELETING_RESULT,
   EXECUTE_PYTHON_ALGORITHM_SUCCESS,
+  CLEAR_RESULT_SUCCESS,
 } from '../shared/types';
 
 const INITIAL_STATE = Map({
```
pyphilia, line 14: Add flag.

hagoptaminian (approved):
looks good to me!

Painguin (changes requested):
Looks good! I've been a lot more minimalist when clearing the state of algorithms (without flags/try-catch). Could you please add the missing things in clearAlgorithm to be consistent.

Painguin (approved):

PR #77, opened 2020-11-13
feat: check dataset against graasp schema

Painguin:
This PR introduces a graasp schema validator with which we can check if a dataset has the expected shape.

For now, the check is made only in the getDataset listener and a `valid` boolean is passed together in the payload.
We could also check the shape when we add/edit a dataset and show which datasets are valid in the `Dataset.js` table or restrict the execution of algorithms in `Executions.js`.

The schema keys that I set as required are the ones that we used in the app (dataset table information and visualizations).
We could be more fine-grained and additionally have a specific schema for each use case. The schemas wouldn't be nearly as big as we would only specify the fields that we need.
Another use would be to have a (small) schema for each algorithm specifying the required fields and check the shape before executing the algorithm.

(The fields that are specified but not required could still invalidate the dataset if the field is present with the wrong type)

Let me know what you think @pyphilia @hagoptaminian.

### DIFF
--- a/src/components/Datasets.js
+++ b/src/components/Datasets.js
@@ -15,15 +15,10 @@ import CodeIcon from '@material-ui/icons/Code';
 import DeleteIcon from '@material-ui/icons/Delete';
 import Tooltip from '@material-ui/core/Tooltip';
 import Alert from '@material-ui/lab/Alert';
-import Chip from '@material-ui/core/Chip';
 import Main from './common/Main';
 import Loader from './common/Loader';
 import LoadDatasetButton from './LoadDatasetButton';
-import {
-  DEFAULT_LOCALE_DATE,
-  SCHEMA_LABELS,
-  SCHEMA_COLORS,
-} from '../config/constants';
+import { DEFAULT_LOCALE_DATE } from '../config/constants';
 import { getDatasets, deleteDataset } from '../actions';
 import { buildDatasetPath, LOAD_DATASET_PATH } from '../config/paths';
 import Table from './common/Table';
@@ -39,6 +34,7 @@ import {
   buildDatasetsListDeleteButtonClass,
 } from '../config/selectors';
 import { SCHEMA_TYPES } from '../shared/constants';
+import SchemaTag from './common/SchemaTag';
 
 const styles = (theme) => ({
   addButton: {
@@ -208,15 +204,9 @@ class Datasets extends Component {
                   {name}
                 </Typography>
               </Grid>
-              {schemaType !== SCHEMA_TYPES.NONE && (
+              {schemaType && schemaType !== SCHEMA_TYPES.NONE && (
                 <Grid item>
-                  <Tooltip title={t('Dataset has Graasp schema')}>
-                    <Chip
-                      size="small"
-                      label={SCHEMA_LABELS[schemaType]}
-                      style={SCHEMA_COLORS[schemaType]}
-                    />
-                  </Tooltip>
+                  <SchemaTag schemaType={schemaType} />
                 </Grid>
               )}
             </Grid>
--- a/src/components/Datasets.js
+++ b/src/components/Datasets.js
@@ -19,7 +19,11 @@ import Chip from '@material-ui/core/Chip';
 import Main from './common/Main';
 import Loader from './common/Loader';
 import LoadDatasetButton from './LoadDatasetButton';
-import { DEFAULT_LOCALE_DATE, SCHEMA_LABELS } from '../config/constants';
+import {
+  DEFAULT_LOCALE_DATE,
+  SCHEMA_LABELS,
+  SCHEMA_COLORS,
+} from '../config/constants';
 import { getDatasets, deleteDataset } from '../actions';
 import { buildDatasetPath, LOAD_DATASET_PATH } from '../config/paths';
 import Table from './common/Table';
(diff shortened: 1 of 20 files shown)


pyphilia (changes requested):
Thank you for this PR!

Here some thought:
- Why not have a generic `detectDatasetSchema` or `detectSchema` and return some types like `GRAASP` or `MOODLE`. It would be easier to add schemas later.
- I would rather run this validation when adding a dataset, then when editing rather than on `getDataset`. What do you think?

src/components/Datasets.js
```diff
@@ -188,22 +192,37 @@ class Datasets extends Component {
...
+                  className={buildDatasetsListNameClass(name)}
+                  variant="subtitle1"
+                  key="name"
+                >
+                  {name}
+                </Typography>
+              </Grid>
+              {schemaType !== SCHEMA_TYPES.NONE && (
+                <Grid item>
+                  <Chip
```
hagoptaminian, line 209: wrap in a Tooltip with text like: _Graasp dataset detected_ or _Dataset has Graasp schema_

src/components/Datasets.js
```diff
@@ -188,22 +192,37 @@ class Datasets extends Component {
...
+                    size="small"
+                    color="primary"
```
hagoptaminian, line 211: Parameterize this similar to the label, as other schemas won't have their `<Chip />` be the same purple color

src/components/dataset/DatasetScreen.js
```diff
@@ -125,22 +136,39 @@ class DatasetScreen extends Component {
...
-                content={datasetContent}
-                id={datasetId}
-              />
+              <Grid container alignItems="center" spacing={2}>
+                <Grid item>
+                  <Typography variant="h5">{t('Content')}</Typography>
+                </Grid>
+                <Grid item>
+                  {datasetSchemaType !== SCHEMA_TYPES.NONE && (
+                    <Chip
```
hagoptaminian, line 145: Wrap in Tooltip like in `/Datasets.js`

hagoptaminian (changes requested):
@Painguin looks really good and works well. I left some small comments. I think it would be nice if we could also see the Graasp label/chip in the Dataset dropdown in the Executions tab. if you think so too, you could add it to this PR

src/components/dataset/DatasetScreen.js
```diff
@@ -125,22 +137,41 @@ class DatasetScreen extends Component {
...
-                content={datasetContent}
-                id={datasetId}
-              />
+              <Grid container alignItems="center" spacing={2}>
+                <Grid item>
+                  <Typography variant="h5">{t('Content')}</Typography>
+                </Grid>
+                <Grid item>
+                  {datasetSchemaType !== SCHEMA_TYPES.NONE && (
+                    <Tooltip title={t('Dataset has Graasp schema')}>
```
pyphilia, line 146: I would rather go for `follows Graasp Schema` but I am not sure maybe @hagoptaminian would know better.

src/components/execution/AddExecutionForm.js
```diff
@@ -124,9 +132,19 @@ class AddExecutionForm extends Component {
 
     const datasetMenuItems = datasets
       .sortBy(({ name }) => name)
-      .map(({ id, name }) => (
+      .map(({ id, name, schemaType }) => (
         <MenuItem value={id} key={id} className={classes.menuItem}>
           {name}
+          {schemaType !== SCHEMA_TYPES.NONE && (
+            <Tooltip title={t('Dataset has Graasp schema')}>
```
pyphilia, line 139: Maybe refactor as a component? With `schemaType` as prop.

src/config/constants.js
```diff
@@ -38,3 +41,15 @@ export const ADD_OPTIONS = {
   FILE: 'file',
   EDITOR: 'editor',
 };
+
+export const SCHEMA_LABELS = {
+  [SCHEMA_TYPES.GRAASP]: 'Graasp',
+  [SCHEMA_TYPES.NONE]: 'Unknown',
```
pyphilia, line 47: `'None'`

src/langs/fr.json
```diff
@@ -74,6 +74,7 @@
     "Save": "Enregistrer",
     "An unexpected error happened while opening the algorithm.": "Une erreur s'est produite lors de l'ouverture de l'algorithme.",
     "your code goes here": "écrivez votre code ici",
-    "You are modifying a Graasp algorithm. Saving will create a new file instead.": "Vous modifiez un algorithme de Graasp. L'enregistrer créera un nouveau fichier."
+    "You are modifying a Graasp algorithm. Saving will create a new file instead.": "Vous modifiez un algorithme de Graasp. L'enregistrer créera un nouveau fichier.",
```
pyphilia, line 77: `Vous êtes en train de modifier` ? Ca me semble plus naturel

pyphilia (changes requested):
Love this tag! I see this is a preset from material so let it be. I let my proposition here (a chip a bit smaller like follows) for if we intend to change it in the future.

I however notice a small bug: there is indeed the check when the dataset file is edited, but the tag is still visible even when the schema does not match. It is only when I return to the Datasets homepage that the chip disapears.

Screenshot 2020-11-17 at 11 11 08

pyphilia (changes requested):
Looks good to me!
Could you also check the schema in Visualisations and add the tag in the list as well? (only Graasp schema datasets can have visualisations)

src/components/charts/AddVisualizationForm.js
```diff
@@ -54,22 +75,27 @@ const AddVisualizationForm = ({
...
-            disabled={!selectedDatasetId}
-            onClick={fetchDatasetToBeVisualized}
-          >
-            {t('Visualize')}
-          </Button>
+          {!selectedDatasetId || isGraasp ? (
+            button
+          ) : (
+            <Tooltip title={t('Only Graasp datasets can be visualized')}>
+              <div>{button}</div>
```
pyphilia, line 96: Is the `div` necessary?

pyphilia (approved):
Looks all good to me now ;)

src/components/charts/AddVisualizationForm.js
Painguin: Yes, otherwise the tooltip doesn't show on disabled buttons

PR #78, opened 2020-11-13
feat: move recurrent python functions into utils file

### LINKED ISSUE #67: Create python utils file
Create a python utils file containing common functions like `save_dataset` and `load_dataset`. Allow the user to edit that file to add its own recurrent little functions.

Painguin:
Moves the recurrent python functions (`save_dataset`, `load_dataset` and `parse_arguments`) into a `utils.py` file. The user can edit that file through the `Algorithms.js` page. 

close #67.


src/components/algorithm/Algorithms.js
```diff
@@ -85,6 +91,12 @@ class Algorithms extends Component {
     push(buildEditAlgorithmPath(id));
   }
 
+  handleUtilsEdit() {
+    const { utils } = this.props;
+    const { id } = utils.first();
```
pyphilia, line 96: Add a todo that we might implement the management of multiple utils files.

src/langs/fr.json
```diff
@@ -74,6 +74,7 @@
     "Save": "Enregistrer",
     "An unexpected error happened while opening the algorithm.": "Une erreur s'est produite lors de l'ouverture de l'algorithme.",
     "your code goes here": "écrivez votre code ici",
-    "You are modifying a Graasp algorithm. Saving will create a new file instead.": "Vous modifiez un algorithme de Graasp. L'enregistrer créera un nouveau fichier."
+    "You are modifying a Graasp algorithm. Saving will create a new file instead.": "Vous modifiez un algorithme de Graasp. L'enregistrer créera un nouveau fichier.",
+    "Edit utils": "Modifier utils"
```
pyphilia, line 78: Would it be too long "Modifier le fichier utils/utilitaire"? Not sure this makes sense aha.

src/reducers/algorithmsReducer.js
```diff
@@ -30,7 +32,23 @@ export default (state = INITIAL_STATE, { type, payload }) => {
         updateActivityList(payload),
       );
     case GET_ALGORITHMS_SUCCESS:
-      return state.setIn(['algorithms'], List(payload));
+      return state
```
pyphilia, line 35: Why not have this separation directly from the back-end in the listener? The payload could look like `{ algorithms: [...], utils: [...] }`

pyphilia (changes requested):
Nice work! Here some comments:

- At some point it will be interesting to have a minimum corrector for algorithm and dataset. Maybe not run the file but at least prevent typos and small errors. It would be nice to have a check for errors in the utils file. Because if this file contains errors, it will fail all other algorithms even though these are correct.
- I see that you are using the same page for edit algo and edit utils. Maybe some checks in the Edit algorithm to display instead "Edit Utils File" and not let the user update the name and description (as it is unnecessary now). Maybe an alert would be interesting to let the user know what this utils file is used for.

public/app/listeners/saveAlgorithm.js
```diff
@@ -22,13 +22,6 @@ const saveAlgorithm = (mainWindow, db) => async (event, { metadata, code }) => {
       type: SAVE_ALGORITHM_SUCCESS,
     });
   } catch (err) {
-    if (err.code === 'ENOENT') {
-      logger.error('File not found!');
-      return mainWindow.webContents.send(SAVE_ALGORITHM_CHANNEL, {
-        type: SAVE_ALGORITHM_ERROR,
-        error: ERROR_MISSING_FILE,
-      });
-    }
```
pyphilia, line 31: Why remove this part?

src/components/algorithm/EditAlgorithm.js
```diff
@@ -163,7 +163,7 @@ class EditAlgorithm extends Component {
     return (
       <Main>
         <Container>
-          <h1>{t('Edit Algorithm')}</h1>
+          <h1>{t('Edit algorithm')}</h1>
```
pyphilia, line 166: Why not capitalized?

src/langs/en.json
```diff
@@ -88,6 +88,9 @@
     "Many users selected...": "{{firstSelection}}, {{secondSelection}}, and {{count}} other selected",
     "Many users selected..._plural": "{{firstSelection}}, {{secondSelection}}, and {{count}} others selected",
     "There was an error visualizing this dataset.": "There was an error visualizing this dataset.",
-    "This dataset does not contain any actions.": "This dataset does not contain any actions."
+    "This dataset does not contain any actions.": "This dataset does not contain any actions.",
+    "Edit utils": "Edit utils",
+    "This file is unmodifiable. Use the user file to write your own functions.": "This file is unmodifiable. Use the user file to write your own functions.",
```
pyphilia, line 93: Would `cannot be modified` be better?

src/langs/fr.json
```diff
@@ -87,6 +87,9 @@
     "Many users selected...": "{{firstSelection}}, {{secondSelection}}, et un autre...",
     "Many users selected..._plural": "{{firstSelection}}, {{secondSelection}}, et {{count}} autres...",
     "There was an error visualizing this dataset.": "Une erreur s'est produite en visualisant ce dataset",
-    "This dataset does not contain any actions.": "Ce dataset ne contient aucune action"
+    "This dataset does not contain any actions.": "Ce dataset ne contient aucune action",
+    "Edit utils": "Modifier le fichier utils/utilitaire",
+    "This file is unmodifiable. Use the user file to write your own functions.": "Ce fichier n'est pas modifiable. Utilisez le fichier utilisateur pour écrire vos propres fonctions.",
```
pyphilia, line 92: `ne peut pas être modifier`?

pyphilia (changes requested):
Looks good! Here some minor comments.

public/app/listeners/saveAlgorithm.js
Painguin: I believe there will never be a missing file error, `fs.writeFileSync` writes into the target location regardless of whether it exists or not.

src/components/algorithm/EditAlgorithm.js
Painguin: I don't know what happened hahaha

pyphilia (approved):

hagoptaminian (commented):
@Painguin very nice. it worked well for me. I really liked the comment in the utils file telling me how to use the function I just wrote ("# You can then use... from utils import function_name"). I feel like we could include more such guidance in the new algorithm window (e.g. explaining in one line: "args = parse_arguments()" does this... "dataset = ..." does that). few suggestions:

1. Add a tooltip to the 'Edit Utils' button: Not everyone will necessarily know what a utils file is. Something like: _You can use the 'utils' file to write functions you want to use across your custom algorithms._

2. At some point it will be nice to add Electron 'accelerators' in the application (i.e. keyboard shortcuts), e.g. when I'm editing the utils file, I instinctively hit Cmd + S to save, and would expect the behavior to work in this editor.

Painguin:
Thank you for your suggestions!
I agree with the 'accelerators' idea. It happens I can also create shortcuts specifically for the editor, so I've added the save shortcut.
I'll also add more informative comments for writing algorithms in a later PR (for handling the algorithm parameters).

PR #98, opened 2020-11-20
feat: add confirmation before delete datasets, results, algo

### LINKED ISSUE #50: Prevent complete deletion
Either add a pop up to ask for confirmation, or implement a trash to restore unwanted deleted items.

### LINKED ISSUE #83: capitalise "Actions" in "Quick actions"

### LINKED ISSUE #90: do not allow to delete Graasp algorithms
It doesn't make much sense to not allow editing, but allow deletion. Better to forbid the deletion of Graasp algorithms.

pyphilia:
close #50 
close #83 
close #90

### DIFF
--- a/src/actions/algorithms.js
+++ b/src/actions/algorithms.js
@@ -8,6 +8,7 @@ import {
   ADD_ALGORITHM_CHANNEL,
   GET_UTILS_CHANNEL,
   SAVE_UTILS_CHANNEL,
+  SHOW_CONFIRM_DELETE_PROMPT_CHANNEL,
 } from '../shared/channels';
 import {
   FLAG_GETTING_ALGORITHMS,
@@ -48,18 +49,25 @@ export const getAlgorithms = () => (dispatch) => {
   }
 };
 
-export const deleteAlgorithm = ({ id }) => (dispatch) => {
+export const deleteAlgorithm = ({ id, name }) => (dispatch) => {
   const flagDeletingAlgorithm = createFlag(FLAG_DELETING_ALGORITHM);
   try {
-    dispatch(flagDeletingAlgorithm(true));
-
-    window.ipcRenderer.send(DELETE_ALGORITHM_CHANNEL, { id });
+    window.ipcRenderer.send(SHOW_CONFIRM_DELETE_PROMPT_CHANNEL, { name });
     window.ipcRenderer.once(
-      DELETE_ALGORITHM_CHANNEL,
-      async (event, response) => {
-        dispatch(response);
+      SHOW_CONFIRM_DELETE_PROMPT_CHANNEL,
+      (e, shouldDelete) => {
+        if (shouldDelete) {
+          dispatch(flagDeletingAlgorithm(true));
+          window.ipcRenderer.send(DELETE_ALGORITHM_CHANNEL, { id });
+          window.ipcRenderer.once(
+            DELETE_ALGORITHM_CHANNEL,
+            async (event, response) => {
+              dispatch(response);
+              getAlgorithms()(dispatch);
+            },
+          );
+        }
         dispatch(flagDeletingAlgorithm(false));
-        return getAlgorithms()(dispatch);
       },
     );
   } catch (err) {
--- a/src/actions/dataset.js
+++ b/src/actions/dataset.js
@@ -6,6 +6,7 @@ import {
   DELETE_DATASET_CHANNEL,
   LOAD_DATASET_CHANNEL,
   SET_DATASET_FILE_CHANNEL,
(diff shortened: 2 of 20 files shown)


src/actions/algorithms.js
```diff
@@ -42,18 +43,25 @@ export const getAlgorithms = () => (dispatch) => {
...
+          window.ipcRenderer.send(DELETE_ALGORITHM_CHANNEL, { id });
+          window.ipcRenderer.once(
+            DELETE_ALGORITHM_CHANNEL,
+            async (event, response) => {
+              dispatch(response);
+              getAlgorithms()(dispatch);
+            },
+          );
+        }
         dispatch(flagDeletingAlgorithm(false));
```
Painguin, line 64: Move `dispatch(flagDeletingAlgorithm(false));` at the end of the `window.ipcRenderer.once` callback

src/actions/dataset.js
```diff
@@ -82,17 +83,27 @@ export const getDataset = ({ id }) => (dispatch) => {
...
+          window.ipcRenderer.send(DELETE_DATASET_CHANNEL, { id });
+          window.ipcRenderer.once(
+            DELETE_DATASET_CHANNEL,
+            async (event, response) => {
+              dispatch(response);
+              getDatasets()(dispatch);
+            },
+          );
+        }
+        dispatch(flagDeletingDataset(false));
```
Painguin, line 104: Move `dispatch(flagDeletingDataset(false));` at the end of the `window.ipcRenderer.once` callback

src/actions/result.js
```diff
@@ -53,17 +54,26 @@ export const getResult = ({ id }) => (dispatch) => {
...
+        if (shouldDelete) {
+          dispatch(flagDeletingResult(true));
+          window.ipcRenderer.send(DELETE_RESULT_CHANNEL, { id });
+          window.ipcRenderer.once(
+            DELETE_RESULT_CHANNEL,
+            async (event, response) => {
+              dispatch(response);
+            },
+          );
+          dispatch(flagDeletingResult(false));
```
Painguin, line 73: Move `dispatch(flagDeletingResult(false));` at the end of the `window.ipcRenderer.once` callback

Painguin (changes requested):
Nice work @pyphilia!
Would there be a way to also translate those prompt messages and maybe also the toastr notifications?

pyphilia:
Good catches! Yes we can, for the prompts probably we will need either to pass the translated text or provide `t`. For toastr I think messages are already wrapped around `t` but the translations are missing.

Painguin (approved):

src/reducers/resultReducer.js
```diff
@@ -36,6 +37,10 @@ export default (state = INITIAL_STATE, { type, payload }) => {
       return state.update('results', (results) => results.push(payload));
     case CLEAR_RESULT_SUCCESS:
       return state.setIn(['current', 'content'], Map());
+    case DELETE_RESULT_SUCCESS:
+      return state.updateIn(['results'], (results) =>
+        results.filter((r) => r.id !== payload),
```
hagoptaminian, line 42: Even though `results` is right there, I think `results.filter((result) => result.id !== payload)` is clearer

src/components/algorithm/Algorithms.js
```diff
@@ -18,6 +18,7 @@ import { getAlgorithms, deleteAlgorithm } from '../../actions';
 import Loader from '../common/Loader';
 import Table from '../common/Table';
 import { buildEditAlgorithmPath, ADD_ALGORITHM_PATH } from '../../config/paths';
+import { AUTHOR_GRAASP, AUTHOR_USER } from '../../config/constants';
```
hagoptaminian, line 21: my linter alerted me to this: `AUTHOR_GRAASP is defined but never used`. not sure why yours didn't catch it!

src/components/algorithm/Algorithms.js
```diff
@@ -152,6 +153,25 @@ class Algorithms extends Component {
...
+      if (author === AUTHOR_USER) {
+        quickActions.push(
+          <Tooltip title={t('Delete algorithm')} key="delete">
+            <IconButton
+              aria-label="delete"
+              onClick={() => this.handleDelete({ id, name })}
+            >
+              <DeleteIcon />
+            </IconButton>
+          </Tooltip>,
```
hagoptaminian, line 172: IMO for visual alignment, it would be better to keep the DeleteIcon there for all algos, but have it disabled for non-author algos. that way the edit icons also line up.

hagoptaminian: oh also this might make sense since we're already using this 'design pattern' in the app (i.e. showing disabled icons). so we stay consistent with that approach

PR #101, opened 2020-11-26
100/GitHub actions

### LINKED ISSUE #100: Replace codeship by github actions
Codeship builds often fail because of some pull rate limit.

Try Github actions as an alternative

Painguin:
Adds a github action file. 

Codeship links/files still have to be deleted. I would need some help/directions here @pyphilia :smile:.  

We can also add slack notifications with [action-slack](https://github.com/marketplace/actions/action-slack) or [Slack Notify](https://github.com/marketplace/actions/slack-notify).

close #100.

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -1,0 +1,20 @@
+name: lint and test
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - name: Install dependencies
+        run: yarn
+      - name: lint
+        run: yarn lint
+      - name: test
+        run: yarn ci
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -1,6 +1,11 @@
 name: lint and test
 
-on: [push]
+on:
+  # trigger workflow on pull requests and merges (merging = pushing on master)
+  pull_request:
+  push:
+    branches:
+      - master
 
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,22 @@
+name: lint and test
+
+on: [push]
```
pyphilia, line 3: I think triggering the actions only inside pull requests and merge events would be nicer. I am not sure how this works but there are these settings https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,22 @@
...
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - name: Install dependencies
+        run: yarn
+      - name: setup
+        run: yarn setup
+      - name: lint
```
pyphilia, line 19: Maybe add prettier check here as well.

.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,22 @@
...
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - name: Install dependencies
```
pyphilia, line 15: I would maybe merge `yarn` and `yarn setup` in the same step.

pyphilia (changes requested):
Thank you for setting this up! Here some comments.

pyphilia (approved):

PR #106, opened 2020-12-04
feat(datasets): preview dataset name when submitting file

### LINKED ISSUE #93: show that the default name of a dataset will be the file name
When adding a dataset, the file name will be the name of the dataset. I didn't notice this, but I was always writing the name of the file manually.

It would be nice if when you select a file, the Dataset Name field is automatically filled with the name of the file.

pyphilia:
close #93

### DIFF
--- a/src/components/Datasets.js
+++ b/src/components/Datasets.js
@@ -89,14 +89,6 @@ class Datasets extends Component {
     push(buildDatasetPath(id));
   };
 
-  handlePublish = () => {
-    // TODO: implement publish functionality
-  };
-
-  handleEdit = () => {
-    // TODO: implement edit functionality
-  };
-
   handleDelete = ({ id, name }) => {
     const { dispatchDeleteDataset } = this.props;
     dispatchDeleteDataset({ id, name });
--- a/src/components/Datasets.js
+++ b/src/components/Datasets.js
@@ -9,8 +9,6 @@ import Grid from '@material-ui/core/Grid';
 import Container from '@material-ui/core/Container';
 import Typography from '@material-ui/core/Typography';
 import IconButton from '@material-ui/core/IconButton';
-import PublishIcon from '@material-ui/icons/Publish';
-import EditIcon from '@material-ui/icons/Edit';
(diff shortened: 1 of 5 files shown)


src/components/Datasets.js
```diff
@@ -252,24 +250,6 @@ class Datasets extends Component {
...
-          </Tooltip>,
-          <Tooltip title={t('Publish dataset')} key="publish">
-            <IconButton
-              disabled
-              aria-label="publish"
-              onClick={() => this.handlePublish(dataset)}
-            >
-              <PublishIcon />
-            </IconButton>
-          </Tooltip>,
```
Painguin, line 272: remove also the functions `handlePublish`, `handleEdit`

src/components/LoadDatasetModal.js
```diff
@@ -53,8 +55,16 @@ class LoadDatasetModal extends Component {
     dispatchAddDataset: PropTypes.func.isRequired,
   };
 
+  buildDefaultDatasetName = (newFileLocation) => {
+    return path.basename(newFileLocation);
+  };
```
Painguin, line 60: We should also remove the `.json` extension from the name.
Also remove the default name logic from the listener `loadDataset.js` since it is done here.

Painguin (changes requested):
Nice. Left two small comments

src/components/LoadDatasetModal.js
```diff
@@ -55,15 +54,11 @@ class LoadDatasetModal extends Component {
...
-  buildDefaultDatasetName = (newFileLocation) => {
-    return path.basename(newFileLocation);
-  };
-
   handleLocationInput = (event) => {
     const fileLocation = event.target.value;
     this.setState({
       fileLocation,
-      defaultDatasetName: this.buildDefaultDatasetName(fileLocation),
+      defaultDatasetName: this.getNameFromLocation(fileLocation),
```
Painguin, line 61: Function `getNameFromLocation` doesn't belong to `this` and should be imported instead

Painguin (changes requested):

Painguin (approved):

PR #107, opened 2020-12-15
96/size limit

### LINKED ISSUE #96: Settings: File size limit to display
Instead of choosing ourselves the limit, set a default file size limite to display and let the user change it in the settings.

+ open file if it is allowed

pyphilia:
close #96 

I added tests, and move SettingsModal as a whole page (because it was difficult to test and also because at some point we would do it).


src/config/constants.js
```diff
@@ -113,3 +111,6 @@ export const UTILS_FILES = {
 };
 
 export const CIRCLE_PROGRESS_SIZE = 26;
+
+export const FILE_SIZE_LIMIT_OPTIONS = [500, 1000, 2500, 5000, 10000];
```
pyphilia, line 115: I put these as tests, let me know which value we should add.

Painguin (changes requested):
Well done @pyphilia. Left minor comments

src/config/constants.js
Painguin: Those look good to me. I'm wondering a infinite/no limit option :thinking:

pyphilia: It would be nice indeed. Let's create an issue with it as "nice to have".

public/app/listeners/getSettings.js
```diff
@@ -0,0 +1,25 @@
...
+const logger = require('../logger');
+const { ERROR_GENERAL } = require('../../shared/errors');
+const {
+  GET_SETTINGS_SUCCESS,
+  GET_SETTINGS_ERROR,
+} = require('../../shared/types');
+
+const getSettings = (mainWindow, db) => async () => {
+  try {
+    const settings = db.get('settings').value();
```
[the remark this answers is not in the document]
pyphilia: I guess there is no reason. I can factor it out.

public/app/listeners/showConfirmOpenDatasetPrompt.js
```diff
@@ -0,0 +1,25 @@
...
+  { name, size },
+) => {
+  logger.debug('showing confirm open dataset prompt');
+
+  const options = {
+    type: 'warning',
+    buttons: ['Cancel', 'Continue'],
+    defaultId: 1,
+    cancelId: 0,
+    message: `The dataset ${name}'s size is ${size} and opening it might take time. Do you want to proceed?`,
```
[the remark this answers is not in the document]
pyphilia: I noticed as well the format, I changed it. Thank you for the "tip" 👍

src/components/Settings.js
```diff
@@ -0,0 +1,177 @@
...
+      }, 0);
+    }
+
+    return (
+      <FormControl
+        className={classes.formControl}
+        id={SETTINGS_FILE_SIZE_LIMIT_SELECT_ID}
+      >
+        <InputLabel shrink htmlFor="fileSizeLimitId">
+          {t('File Size Limit')}
```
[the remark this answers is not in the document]
pyphilia: I changed a little bit the layout of settings to allow a description instead.
Screenshot 2021-01-12 at 12 17 46

src/components/dataset/ViewDatasetButton.js
```diff
@@ -0,0 +1,63 @@
+import React, { Component } from 'react';
```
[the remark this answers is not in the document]
pyphilia: Good point, I will add it for results as well.

public/electron.js
```diff
@@ -414,6 +427,15 @@ app.on('ready', async () => {
...
   ipcMain.on(SAVE_UTILS_CHANNEL, saveUtils(mainWindow));
 
+  // called when setting file size limit
+  ipcMain.on(SET_FILE_SIZE_LIMIT_CHANNEL, setFileSizeLimit(mainWindow, db));
+
+  // called when gettings file size limit
+  ipcMain.on(GET_FILE_SIZE_LIMIT_CHANNEL, getFileSizeLimit(mainWindow, db));
+
+  // called when gettings settings
+  ipcMain.on(GET_SETTINGS_CHANNEL, getSettings(mainWindow, db));
```
Painguin, line 437: ~gettings~getting

src/components/Settings.js
```diff
@@ -0,0 +1,184 @@
...
+
+    return (
+      <FormControl
+        className={classes.formControl}
+        id={SETTINGS_FILE_SIZE_LIMIT_SELECT_ID}
+      >
+        <FormLabel>{t('File Size Limit')}</FormLabel>
+        <FormHelperText>
+          {t(
+            'A warning will appear when opening datasets bigger than this limit.',
```
Painguin, line 129: add translation in `en.json` and `fr.json`

Painguin (approved):
It is looking good and works nicely! 

I don't know how much emphasis we put in translations as we develop features, but the pop-up warning should also be translated.

pyphilia:
Indeed we should translate the prompts... I think there are some possibilities for using `i18n` and the `shared` folder https://www.christianengvall.se/electron-localization/
In any case all prompts need to be translated, so I will create an issue including all of them.

PR #109, opened 2020-12-16
feat: add folder indications and copy in add dataset

### LINKED ISSUE #108: Add folder location in each page
+ Add folder location in each page
+ add indication when loading datasets that it will create a copy.

pyphilia:
close #108 

I added the notice in Datasets, Results and Algorithms.

Screenshot 2020-12-16 at 15 10 43

In Import/Add new Dataset modal:

Screenshot 2020-12-16 at 15 10 53

I also took this opportunity to refactor a bit the constants, to split paths and common constants in `public`


src/components/algorithm/Algorithms.js
```diff
@@ -61,6 +62,13 @@ const styles = (theme) => ({
   },
   utilsButton: {
     float: 'right',
+    position: 'absolute',
+    right: theme.spacing(2),
+    top: DRAWER_HEADER_HEIGHT + theme.spacing(3),
+  },
```
Painguin, line 68: The button seems to be overlapping with the alert if the window is not big enough:
image

I think you wanted to put the button in the far top right. You could let the float style (with some margin) and move the button outside of the `Container` so that it isn't constrained by the container area

Painguin (approved):
Looking good :+1:

hagoptaminian (commented):
nicely done. two thoughts:

- maybe give an option to make these Alerts closable, like here: https://material-ui.com/components/alert/#TransitionAlerts.js

- and/or it would be nice if there were a quick action that opens these files in their folders. just today I used a similar feature in vscode (below) and it is common in many apps:

Screen Shot 2021-01-11 at 12 16 54 PM

but it looks very  good in any case. let me know what you think.

pyphilia:
I'm am totally agreeing with the closable option. However if it is included I would add settings to keep it close (otherwise it will show each time you come back on these pages) as well as a switch (or something else) in the settings to show it again. And since it is only a `nice to have` I will add it as a new issue.
I will work on opening the corresponding folder and update this PR.

src/components/algorithm/Algorithms.js
pyphilia: I moved it under.

Painguin (approved):

Painguin (approved):

PR #121, opened 2021-01-27
feat: add shuffle algorithm

Painguin:
The shuffle algorithm takes a field selection parameter to select the attributes to shuffle. Although any field can be selected, only the ones that are children to arrays will be shuffled.

two potential flaws:
- The dependencies between the shuffled attributes and the others are broken
- each user has multiple records. So if a shuffled attribute has the same value for every record of a user, an attacker might count the occurrences and link the user back

### DIFF
--- a/public/app/config/graaspAlgorithms.js
+++ b/public/app/config/graaspAlgorithms.js
@@ -146,7 +146,8 @@ const GRAASP_ALGORITHMS = [
   {
     id: 'shuffle-fields',
     name: 'Shuffle fields',
-    description: 'Shuffle selected array fields from a dataset',
+    description:
+      "Given an array of objects, this algorithm randomly shuffles the values in selected keys between the array's objects",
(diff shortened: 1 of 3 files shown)


pyphilia (approved):

public/app/config/graaspAlgorithms.js
```diff
@@ -143,6 +143,40 @@ const GRAASP_ALGORITHMS = [
       },
     ],
   },
+  {
+    id: 'shuffle-fields',
+    name: 'Shuffle fields',
+    description: 'Shuffle selected array fields from a dataset',
```
hagoptaminian, line 149: 'Given an array of objects, this algorithm randomly shuffles the values in selected keys between the array's objects.'

A bit of a mouthful, but...!

hagoptaminian (approved):

PR #122, opened 2021-01-27
119/build fail

### LINKED ISSUE #119: build fails on fieldSelectorUnselectAll
Instruction `yarn build` fails:

```
yarn run v1.22.10
$ yarn prepare && env-cmd -f ./.env.local node scripts/setup.js
$ mkdirp public/shared && cpy "src/shared/**/*.js" "public/shared"
$ env-cmd -f ./.env.local react-scripts build
Creating an optimized production build...
Failed to compile.

./src/components/parameter/FieldSelector.js
Attempted import error: 'fieldSelectorUnselectAll' is not exported from '../../shared/utils'.


error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Caused by:

- Object deconstruction: `{...schema, a,b,c}`)
- Array deconstruction in map: `Object.entries(obj).map(([key, value]) => ...)`

in `.../shared/utils` because the operations are not supported when building.

Also write tests for `generateFieldSelector` and `fieldSelectorUnselectAll`.

Painguin:
close #119

### DIFF
--- a/public/shared/utils.js
+++ b/public/shared/utils.js
@@ -13,61 +13,56 @@ const generateFieldSelector = (schema, expandUntilDepth = 1) => {
     switch (type) {
       case 'object':
         // if we have an object, the 'properties' field contains the information for each key
-        return {
-          ...field,
+        return Object.assign(field, {
           selected: false,
           expanded,
           properties: Object.fromEntries(
-            Object.entries(properties).map(([key, value]) => [
-              key,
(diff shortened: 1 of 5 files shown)


pyphilia (approved):
Nice debug 💯 
For more context: the `shared/utils.js` file was using ES6 functionalities, which were not correctly compiled when building the application.

PR #127, opened 2021-02-08
92/hashed users visualization

hagoptaminian:


Painguin (approved):
Looking good!

src/utils/charts.js
```diff
@@ -181,6 +183,16 @@ export const formatActionsByVerb = (actionsByVerbObject) => {
 // 'actions' is an array in the format retrieved from the API: [ { id: 1, user: 2, ... }, {...} ]
 // therefore note: id is the id of the action, and user is the userId of the user performing the action
 export const filterActionsByUser = (actions, usersArray) => {
+  // if the dataset passed into the visualizer has hashed users, then user names are no longer available
+  // instead, users key (and usersArray here) will be of the format... users: {[_id: hashedId], [_id: hashedId2], ...}
+  // therefore slightly different filtering function
+  if (usersArray.some((user) => !user.name)) {
+    return actions.filter((action) => {
+      return usersArray.some((user) => {
```
pyphilia, line 191: I would suggest this structure
```
const userIdsArray = usersArray.map(({_id})=>_id)

const filteredActions = actions.filter((action) => {
      return userIdsArray.includes(action.user);
});
    
```

pyphilia (approved):
Nice! Just a minor comment

src/utils/charts.js
hagoptaminian: You're right, this is a nicer and clearer syntax.

hagoptaminian: Also, you're right. We should be able to select results to visualize from here... although currently we only visualize datasets with the Graasp schema, so if an algorithm has altered a dataset such that it no longer matches the Graasp schema, we won't be able to visualize it. I'll add it as a separate issue.

hagoptaminian:
(@juancarlosfarah tagged you here only for the readme - no need to review the other code)

juancarlosfarah (approved):
🔭

PR #128, opened 2021-02-08
feat: add Schemas tab

### LINKED ISSUE #123: Multiple schema support
Support the addition of other schemas.

Painguin:
- Adds a `Schemas` tab, a `View Schema` screen and a `Add Schema` button.
- Schemas can be visualized, edited and deleted (The Graasp schema cannot be edited or deleted).
- The detected schemas are set automatically on the datasets. Each dataset can have multiple schemas. The list of schemas for each dataset updates automatically as new schemas are added/edited/deleted.
- The `Add Schema` button features the automatic generation of a schema from a dataset. No field from a generated schema will be set as `required` because it is never obvious which fields are mandatory. Therefore, all datasets will match the newly generated schema. But the `required` fields can be added manually when editing the schema. We could maybe offer some guidance for editing the `required` fields.
- The schemas allow the use of algorithms that depend on field selection parameters on datasets of any shape.

TODO:
- [x] Write translations
- [x] Write tests

close #123


src/reducers/schemaReducer.js
```diff
@@ -10,9 +16,28 @@ const INITIAL_STATE = Map({
 export default (state = INITIAL_STATE, { type, payload }) => {
   switch (type) {
     case FLAG_GETTING_SCHEMAS:
-      return state.update('activity', updateActivityList(payload));
+    case FLAG_SETTING_SCHEMA:
+    case FLAG_DELETING_SCHEMA: {
+      if (payload) {
+        return state.update('activity', (list) => list.push(type));
```
pyphilia, line 22: Another solution I could suggest is to have 
`activity: { [FLAG_DELETING_SCHEMA]: [], ... }` and use the same `updateActivityList` function to update it. Then if you need this activity you can flatten the `activity` map and check for a `true` value.

src/utils/color.js
```diff
@@ -0,0 +1,6 @@
+const generateTextColorFromBackground = (rgb) =>
+  rgb[0] * 0.299 + rgb[1] * 0.587 + rgb[2] * 0.114 <= 186 ? 'white' : 'black';
```
pyphilia, line 2: Interesting but very useful xD

src/config/constants.js
```diff
@@ -112,3 +112,8 @@ export const PARAMETER_TYPES_DEFAULT = {
   [PARAMETER_TYPES.INTEGER_INPUT]: 0,
   [PARAMETER_TYPES.STRING_INPUT]: '',
 };
+
+export const DEFAULT_TAG_STYLE = {
+  backgroundColor: '#555555',
+  color: 'white',
```
pyphilia, line 118: Is it possible to use `generateTextColorFromBackground` and get the background from `theme`?

test/unit/shared/utils.js
```diff
@@ -1,29 +1,39 @@
 import { expect } from 'chai';
+import { setFieldSelectorAttributesData } from '../../fixtures/unit/shared/utils';
```
pyphilia, line 2: Test filename should end with `.test.js`

pyphilia (changes requested):
Looks really nice! I'll update this comment once I tested it.

package.json
```diff
@@ -59,6 +60,7 @@
     "redux-devtools-extension": "2.13.8",
     "redux-promise": "0.6.0",
     "redux-thunk": "2.3.0",
+    "styled-components": "5.2.1",
```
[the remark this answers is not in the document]
Painguin: It is a dependency for `material-ui-color`

public/app/listeners/setSchema.js
```diff
@@ -0,0 +1,75 @@
...
+      const dataset = db
+        .get(DATASETS_COLLECTION)
+        .find({ id: fromDataset })
+        .value();
+      const { filepath } = dataset;
+      const content = fs.readFileSync(filepath, 'utf8');
+      const json = JSON.parse(content);
+      schemaDef = generateSchemaFromJSON(json);
+    } else if (!schemaDef) {
+      schemaDef = { type: 'object', properties: {} };
```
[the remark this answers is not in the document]
Painguin: Yes, it corresponds to an empty dataset and the user can build up on it. It is set when the app user selects None for the `From dataset` value

public/app/listeners/setSchema.js
```diff
@@ -0,0 +1,75 @@
...
+    datasets.forEach(({ id: dId, filepath, schemaIds: schemaIdsPrev }) => {
+      const content = fs.readFileSync(filepath, 'utf8');
+      const json = JSON.parse(content);
+      const satisfiesSchema = validateSchema(json, schemaDef);
+
+      if (satisfiesSchema && !schemaIdsPrev?.includes(id)) {
+        // add schema id to schemaIds
+        db.get(DATASETS_COLLECTION)
+          .find({ id: dId })
+          .assign({ schemaIds: [].concat(schemaIdsPrev, id) })
```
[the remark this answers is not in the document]
Painguin: I wanted it to still work in case schemaIdsPrev was undefined. But I realise that it would just become [undefined, id]... I will change this

public/app/schema/detectSchemas.js
```diff
@@ -0,0 +1,22 @@
...
+
+const ajv = new Ajv();
+
+const validateSchema = (json, schema) => {
+  const validator = ajv.compile(schema);
+  return validator(json);
+};
+
+const detectSchemas = (json, db) => {
+  const schemas = db.get(SCHEMAS_COLLECTION).value();
```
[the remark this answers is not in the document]
Painguin: It is indeed possible 😮

src/components/schema/AddSchemaButton.js
```diff
@@ -0,0 +1,59 @@
...
+          onClick={handleClickOpen}
+          className={classes.addButton}
+        >
+          <AddIcon />
+        </IconButton>
+      </Tooltip>
+      <AddSchemaModal
+        open={open}
+        setOpen={setOpen}
+        handleClickOpen={handleClickOpen}
```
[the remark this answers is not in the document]
Painguin: In the mess of "copy-pasting" from other button-modal duos I noticed that some handlers passed to components were not used there. So after cleaning that I went for `handleButtonOnClick` for the `IconButton` onClick handler and `handleModalOnClose` for when the modal closes.

src/reducers/schemaReducer.js
Painguin: I love it! I feel like every reducer should use that logic.

hagoptaminian (commented):
@Painguin this is working really nicely and looking great. I am able to add a dataset, detect and add its schema, and have similar datasets tagged correctly. the parameterized algorithms work correctly on the new added schema. it's a really nice feature and an important step forward in making the app functional with other datasets, and the schema tags look nice and slick.

I am facing primarily one issue/potential bug. Let's say I add a file `exams.json` that is just `{"users": [//some objects], "aggregateScore": 100}`, and add its schema, calling it Exams. Then I add another file `games.json` that is say just `{"totalPoints": 999}`. That latter dataset  (and really all other loaded datasets) is being tagged as Exams as well, which I feel just doesn't make sense.

Of course one way to resolve this is by making `users` and/or `aggregateScore` _required_ in the first schema. Then only datasets that have those two keys will be tagged as Exams. But I am wondering if there is something else we can do? Maybe make the root-level keys in a dataset _required_ by default? Of course, that may cause issues as well (e.g. now someone adds a dataset that they are expecting to be identified as schema X but because it is missing some root level key it is not working), but I feel like it's the more conservative option. 

let me know what you think/let's discuss.

pyphilia:
@hagoptaminian This issues totally makes sense. And this is bound to happen. Your solution is also totally valid, I also  thought about another one (I suggested while talking with @Painguin ) , which is to have the possibility to generate schemas from multiple datasets at the same time. It  would enable the detection of required fields more accurately in that case.

Painguin:
Thank you for your feedback and suggestions!
I agree with the concerns you have with the required fields. In my opinion there is no way to automatically and accurately infer the required fields as there might be a lot of variations across datasets for a same schema. But we can approximate and both of your suggestions are good. I think I would implement the @hagoptaminian's suggestion in this PR so that we avoid schemas matching every single dataset. And then we can think of and implement @pyphilia's idea for another PR.
We can also put more emphasis in the contribution of the user for setting the required fields as he might ultimately be the one who knows best what they should be. As suggested, I've added a [link](https://json-schema.org/learn/getting-started-step-by-step.html) in the app to help the user.

hagoptaminian:
This looks good to me! Very nicely done 👍 

I will add one issue/bug I noticed separately since it might need some work.

src/components/schema/SchemaView.js
```diff
@@ -0,0 +1,309 @@
...
+  title: {
+    marginBottom: theme.spacing(2),
+  },
+  tagWrapper: {
+    marginTop: theme.spacing(2),
+    marginBottom: theme.spacing(2),
+    textAlign: 'center',
+  },
+  tag: {
+    fontSize: 2 * theme.typography.fontSize,
```
pyphilia, line 50: It's possible that `2em` will be enough, but not sure (basically it's 2 times the inherit font size)

pyphilia (approved):
Nice! I left some minor comments here^^

PR #129, opened 2021-02-09
118/disclaimer

pyphilia:
It was a screenshot from a previous build, instead of "Here is my own description" there is "License and Agreements"

Screenshot 2021-02-09 at 12 32 31

image

I don't know how licenses work for linux OS, could you check @Painguin ? Is there an installer like Mac and Windows or is it just some description on a website? Here the documentation https://www.electron.build/configuration/linux


hagoptaminian:
@pyphilia looking great! Instead of _License and Agreements_, can we use _Disclaimer_? It's technically not a license agreement, but rather an additional warning and disclaimer.

pyphilia:
@hagoptaminian No problem. I can change it for dmg, but the Windows installer is a more tricky, so I think it is better if it stays as "License Agreement" in the header.

Painguin (commented):
On linux, using the appImage, the following "disclaimer" is shown:
image

The text stops at that position because it is using the `license.txt` file which includes weird (unreadable) characters. In order to use the `license_en.txt` file instead, you can add the following in `package.json` under the `build` key:
```json
    "appImage": {
      "license": "license_en.txt"
    }
```

Here is the disclaimer with the change: 
image

As a side note, I tried installing and running the `.snap` package. Python was not recognized and a weird behavior happened when I tried loading a dataset from my computer: 
image

hagoptaminian (approved):

pyphilia:
I updated as you proposed Pedro. I also changed `license.txt` to `windows_license.txt` to emphasize the fact it is a windows-only file.

PR #141, opened 2021-02-25
feat: add built-in algorithm

### LINKED ISSUE #138: add button to reload built-in algorithms

Painguin:
Adds an option to add built-in algorithms in the `AddAlgorithm` page.

Closes #138


pyphilia (changes requested):
Thanks for the PR! I left some feedback ;)

- [ ]  Would it be clearer to name these special algorithm as `Graasp Algorithms` instead of `built-in Algorithms`? But we can imagine that we would later provide non-Graasp algorithm such as Moodle algorithm... but I guess there is a difference between OUR Graasp algorithms and algorithms FOR Graasp
- [x] Add a tooltip when adding a built-in algorithm to precise that the code below cannot be modified. Maybe change the theme as well? https://securingsincity.github.io/react-ace/ I see `kuroir` may bring the `disabled` feeling 
-  I honestly think we will need to refactor the process and layout for adding algorithm, as it became really complicated aha But it should be tackled in another PR (by using tabs or steps)
- [x] I see that you cannot add twice a Graasp Algorithm. I don't know if it is really smart or a bug. I would honestly expect to have two identical algorithms when adding twice an algorithm. Plus we have a success feedback when adding one, so I think there is a contradiction. Maybe we should discuss to decide this. (+ you still have the possibility to add an algorithm with the select option even though it is already available)  
- [ ] I am wondering if we could refactor`addGraaspAlgorithm` in `db.js` and `listeners/addAlgorithm.js` I see a lot of similarity maybe an utility function to generally add algorithm would be nice.
- [ ] Maybe add simple tests!

hagoptaminian (commented):
Nicely done @Painguin. It works very well. 

- I agree with @pyphilia that in general this workflow will need to be re-thought. Maybe in the future we can simply rely on the _Load algorithm from file_ functionality, and the algorithms for a specific schema will be stored/shared/downloaded externally. For now though I think this does what we need well.
- Agree also that a tooltip in the 'built-in algorithms' page would really help. Suggested text: _Built-in algorithms are default algorithms provided by the application developer. These algorithms cannot be edited._
- Since there is now a button to clear all data, I think it makes sense that you've made the default algorithms delete-able one by one as well.
- Can we please add a way to also restore the default schema? I am worried that in the next round of testing the 'clear database' button will be clicked, and then there will be no way to recover the schema. Maybe you can simply have a 'Restore application defaults' in the Settings tab, with text that says something like _This restores the application's default data schema and algorithms._ I know this diverges from the current approach but just a suggestion and could help simplify. I am OK with both options.

Good stuff

pyphilia:
I had to big and a lot of changes. I'll try to summary all of them:
- I added a checkbox to allow the user to wipe his database and use default algorithms and the graasp schema
- Some parts of the code and tests requiring default algorithms or the defaults schemas were failing, so I patch everything I could find. 
- Added some indications when adding an algorithm but a required schema is missing.
- Added tests for clearing the database and the algorithm parameters

More technical points:
- move `ALGORITHMS_COLLECTION`, etc in the `shared` folder, which lead to changing all the listener files.
- I factored out some utils from `addAlgorithm` or `clearDatabase` for instance. This allowed me to reuse the functionalities. Particularly, I used the `clearDatabase` process in `setSampleDatabase`.

hagoptaminian:
Nice work @pyphilia! Overall it's working really well: I added datasets, ran algorithms, ran algorithms on the results, exported the data, visualized, etc. All worked as expected... with one exception below.

A few comments:

- At application initialization, the default algorithms and schema are not loaded. Was this intended? I think it makes to have them there by default, since the only other way to reload them would be to clear the database, which is not an intuitive step for an already empty application.
- The main errors I have encountered have to do with schema. Specifically, say you add a random (non-Graasp) dataset, and generate a schema from it. If you then look at that schema, it is empty. Additionally, if you try to apply an algorithm to that dataset in Executions, the application crashes. So something seems to have gone wrong here.
- A thought: Consider removing 'Add default algorithm' from the 'Add algorithm' page. I think it's just an additional, potentially confusing option.
- In Settings, instead of 'Use Sample Dataset', I would strongly recommend something like _Reload Default Algorithms and Schema_, and in the tooltip: _Provided by the application developer, these algorithms and schema are optimized to process Graasp datasets._

PR #153, opened 2021-03-20
Write tests for Pipeline pages and implement add,save and delete pipeline .

### LINKED ISSUE #151: Write tests for Pipeline pages and implement add,save and delete pipeline .

badrlarhdir:
Features implemented : 
- Adding a pipeline from `AddPipeline.js`
- Saving a pipeline from `EditPipeline.js`
- Deleting a pipeline from `Pipelines.js`.

- When saving/adding/deleting a pipeline show a popover to display the message of success/failure of the action. 
- Testing : 
    - `Pipeline` page charging data from `db.json`
    - `Pipeline` page edit/delete functionalities 
    - `AddPipeline`page adding a new pipeline to `db.json`
    - `EditPipeline` page editing a pipeline from `db.json`
  
close #151


pyphilia (changes requested):
Very nice! It works well. I left some comments, and looking forward the missing features ;)

src/components/pipeline/AddPipeline.js
```diff
@@ -26,17 +26,23 @@ const styles = (theme) => ({
...
       <Container className={classes.content}>
         <Typography variant="h4">{t('Add Pipeline')}</Typography>
         <PipelineForm
+          id=""
           pipelineAlgorithms={[]}
           name=""
           description=""
-          handleSave={() => {}}
+          handleSave={false}
+          handleHistory={push}
```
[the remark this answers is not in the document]
badrlarhdir: I think `useHistory `comes from the Routing in `App.js` but the `pipelineForm` is called from a component making `withRouter ` useless to pass this non-existent prop to `PipelineForm`.

src/components/pipeline/PipelineAccordion.js
```diff
@@ -55,19 +110,28 @@ const PipelineAccordion = ({ pipelineAlgorithms }) => {
                 <ListItemIcon>
                   <CodeIcon />
                 </ListItemIcon>
-                <Typography variant="subtitle1">{algorithm.name}</Typography>
+                <Typography variant="subtitle1" id={`name${i + 1}`}>
+                  {applicationAlgorithms.find((x) => x.id === algo.id).name}
```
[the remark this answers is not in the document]
badrlarhdir: For the ids simplification, it depends on if you would allow multiple times the same algorithm to occur in a single pipeline.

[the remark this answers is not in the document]
pyphilia: That's true, so let's keep it as it is.

test/pipelines/pipelines.test.js
```diff
@@ -0,0 +1,88 @@
...
+
+      await addAlgorithmToPipeline(client, FIRST_ALGORITHM);
+      await addAlgorithmToPipeline(client, SECOND_ALGORITHM);
+      await addAlgorithmToPipeline(client, THIRD_ALGORITHM);
+
+      await clickSavePipelineButton(client);
+      await checkPipelineRowLayout(client, PRIMARY_PIPELINE);
+
+      const nbrPipelinesAfterAdd = await getNumberOfPipelines(client);
+      expect(nbrPipelinesAfterAdd - nbrPipelinesPrev).to.be.equal(1);
```
pyphilia, line 70: I see you reused how algorithms where tested. This line was defined as such because we previously had a non-defined number of default algorithms at the start of the app (now it is no more the case). In your case you don't particularly need to test the difference before and after an action, since you define the default data. But you can leave it as it is, it's not a problem.

test/pipelines/utils.js
```diff
@@ -0,0 +1,136 @@
...
+  const matchedAlgorithm = await client.$(
+    `#${PIPELINE_TABLE_ID} .${buildPipelineRowClass(name)}`,
+  );
+
+  expect(
+    await (await matchedAlgorithm.$(`.${PIPELINE_NAME_CLASS}`)).getText(),
+  ).to.equal(name);
+};
+
+export const getNumberOfAlgorithms = async (client) => {
```
pyphilia, line 53: Do you really need this?

test/pipelines/utils.js
```diff
@@ -0,0 +1,136 @@
...
+      name,
+    )} .${PIPELINE_EDIT_BUTTON_CLASS}`,
+  );
+
+  await editPipelineSelect.click();
+};
+
+export const removeAlgorithmPipeline = async (client, algorithmIndex) => {
+  const expandAccordionAlgorithmSelect = await client.$(
+    `#panel${algorithmIndex + 1}-accordion`,
```
pyphilia, line 97: You should create an id builder (like `buildDatasetsListNameClass`) and apply it here and to the corresponding place in the component.

test/pipelines/utils.js
```diff
@@ -0,0 +1,136 @@
...
+  );
+  await expandAccordionAlgorithmSelect.click();
+
+  const removeAlgorithmSelect = await client.$(
+    `#button${algorithmIndex + 1}-remove`,
```
pyphilia, line 102: Same as above.

test/fixtures/pipelines/pipelines.js
```diff
@@ -0,0 +1,42 @@
...
+
+export const FIRST_ALGORITHM_INDEX = 0;
+export const SECOND_ALGORITHM_INDEX = 1;
+export const THIRD_ALGORITHM_INDEX = 2;
+
+export const FIRST_ALGORITHM = {
+  id: 'exec-fast-algo',
+  name: 'algorithm 1',
+  description: `description`,
+  filepath: path.join(__dirname, './executionFast_algorithm.py'),
```
pyphilia, line 21: Maybe let's not duplicate the python algorithms, so comfortably use the file from `fixtures/executions`. Only If you need specific algorithms for your pipelines you can create some.

pyphilia (changes requested):
Nice work! I left you some comments, I think we will soon be able to wrap up this second PR!

test/pipelines/pipelines.test.js
```diff
@@ -0,0 +1,88 @@
...
+      // edit pipeline
+      await clickEditPipelineButton(client, PRIMARY_PIPELINE);
+      await removeAlgorithmPipeline(
+        client,
+        ALGORITHMS_PIPELINES.findIndex((x) => x.id === SECOND_ALGORITHM.id),
+      );
+      await editNamePipeline(client, SECONDARY_PIPELINE);
+      await editDescriptionPipeline(client, SECONDARY_PIPELINE);
+      await clickSavePipelineButton(client);
+
```
pyphilia, line 78: I thing something important to check here is whether the changes are persisted. Eg if you quit the  screen and comeback are the changes actually saved.

test/pipelines/pipelines.test.js
```diff
@@ -0,0 +1,88 @@
...
+      const nbrPipelinesAfterEdit = await getNumberOfPipelines(client);
+      expect(nbrPipelinesAfterEdit - nbrPipelinesAfterAdd).to.be.equal(0);
+
+      // delete pipeline
+      await clickDeletePipelineButton(client, SECONDARY_PIPELINE);
+      const nbrPipelinesAfterDelete = await getNumberOfPipelines(client);
+      expect(nbrPipelinesAfterDelete - nbrPipelinesAfterEdit).to.be.equal(-1);
```
pyphilia, line 85: Something which can actually happen is if you delete a pipeline and it was another one which gets deleted (some goes for edition). Maybe add two pipelines and check if changes are happening on the right pipeline.

pyphilia (changes requested):
Nice! I think we will be able to merge this PR after this iteration^^

test/pipelines/pipelines.test.js
pyphilia: And check if the added algorithms are in the correct order.

pyphilia (approved):

PR #158, opened 2021-04-15
feat: add validation functionality

### LINKED ISSUE #131: custom schema and algorithms: handle .json files whose outermost value is an array
custom algorithms like `suppress fields` expect json datasets to have outermost values that are objects. if a json file like the following is loaded:

```
[
  {
    "name": "x"
  },
  {
    "name": "y"
  }
]
```

and then a schema for it generated, then selecting an algorithm like `suppress fields` crashes the application

### LINKED ISSUE #154: Add dataset validation

Painguin:
Adds a validation page that allows the user to run validation algorithms.

Main points:
- The validation algorithms finish with either a success, warning or failure, together with some informative text.
- 4 validation algorithms are proposed:
  - `Verify potentially dangerous attributes` (graasp only)
  - `Detect users`
  - `Verify k-anonymity`
  - `Verify l-diversity`
- One can filter the algorithms in the `Algorithms` page by type (anonymization or validation (eventually visualization later)).
- Adding and editing algorithms also works for validation algorithms. A type selection has been added to differentiate. There is one distinct template code for each type and a set template button was added.
- When dealing with schemas/field selectors, we don't assume that the root level is an object. So top-level arrays should now be supported.

close #154
close #131


pyphilia (changes requested):
So huge!!
Here some comments:
- Do not allow to validate a dataset if no validation algorithm is available (currently you can click on the + button but the validation select is empty)
- Disabled the square + button as long as no validation algorithm is selected 
- Add the validation algorithms when clicking on "load Graasp Algorithms and schema" (in settings). Maybe this comment is not valid, I just noticed the type select at the top of the page. But I think it would be clearer to have all algorithms available but add a `type` column in the table to allow to order them. We can discuss.
- I got an error when running `Verify k-anonymity` and `Detect users` on Fonctions (will give you the file). Additionally, the loader only updates when quitting the page (I got the error in the terminal but a loader was still showing).
- In the validation status table, I would maybe have a `Status` column containing the loader/result icon and the stop button. Currently the stop button is not aligned depending on the algorithm name length.
- Move the Validation menu just below the Executions one?
- I would add a View button for each line of the Validation Execution to see more details. But this would be another PR.
- This Add Algorithm page is so crowded aha. I can understand why you have this `set default template` button but I feel it should be apply by default when changing the algorithm type. Why not have an alert dialog stating `The algorithm template will be reset and any changes will be lost. Do you confirm?` and apply the template. Honestly this will change when the layout will be refactored, but it seems more viable for me that way. Additionally the `Set Default Template` does not reset the parameters, which is a bug.
- I think you should save the validation executions in `executions` and reference their id in `validation`. You already added a type, so it will be easy to not display them in the Executions Table. Let me know if something is blocking you from doing so. Merging them should let you use the same channels and listeners for normal executions and validation executions (with some tweaks).
- I would maybe extend the validation algorithm by default, once added so that edit parameters is visible.
- Just an idea I got, it would have been fun maybe to display the add validation page with all validation algorithm visible (each one is a squared button) and enable/disable by clicking on them. A clicked box would then display the `edit parameters` button. Just an idea, we can discuss and open an issue for that.

src/components/algorithm/Algorithms.js
```diff
@@ -272,10 +283,16 @@ class Algorithms extends Component {
             {description}
           </Typography>,
         ],
-        author: (
+        type,
+        typeField: (
```
pyphilia, line 287: I suppose you added these to add a class. I don't think such complicated code should be added only for test sake. It will be a bit more complex for the tests, but maybe you can either check a string over a whole row or detect the correct column depending on the position of the header. We can discuss about it.

src/components/common/Table.js
```diff
@@ -77,13 +77,19 @@ const Table = (props) => {
...
             <TableRow key={key} className={className}>
-              {columns.map(({ field, alignField }) => {
-                return (
-                  <TableCell align={alignField} key={field}>
-                    {row[field]}
-                  </TableCell>
-                );
-              })}
+              {columns
+                .filter(({ field }) => field)
```
pyphilia, line 81: Is there a case where `field` is undefined? Won't it shift the columns if you filter it?

src/components/execution/ExecutionTable.js
```diff
@@ -249,7 +249,9 @@ const mapStateToProps = ({ dataset, algorithms, executions, result }) => ({
     Boolean(dataset.getIn(['activity']).size) &&
     Boolean(algorithms.getIn(['activity']).size) &&
     Boolean(executions.get('activity').size),
-  executions: executions.getIn(['executions']),
+  executions: executions
+    .getIn(['executions'])
+    .filter(({ type }) => type === ALGORITHM_TYPES.ANONYMIZATION),
```
pyphilia, line 254: This check is weird (use an algorithm type to check an execution type). We should either have separate types for executions or `type` should be contained in the `algorithm` key.

src/components/parameter/FieldSelector.js
```diff
@@ -42,28 +43,18 @@ const FieldSelector = ({ schema, onChange, disabled, id }) => {
   const { t } = useTranslation();
   const classes = useStyles();
 
-  if (!schema || !schema.properties) {
+  if (!schema) {
```
pyphilia, line 46: What if `properties` is null?

src/components/parameter/FieldSelector.js
```diff
@@ -110,8 +98,69 @@ const FieldSelectorTree = ({ name, field, onChange, disabled }) => {
...
+          ),
+        );
+
+        Array.prototype.push.apply(children, newChildren);
+        childSelected =
+          childSelected || anySelected({ type: 'object', properties });
+      }
+    }
+  }
+
```
pyphilia, line 161: Maybe add some comments for these lines.

src/components/parameter/FieldSelector.js
```diff
@@ -129,54 +178,26 @@ const FieldSelectorTree = ({ name, field, onChange, disabled }) => {
...
-                }}
-                disabled={disabled}
-              />
-            );
-          })}
-        </div>
+      {
+        <FormControlLabel
+          control={
+            // eslint-disable-next-line react/jsx-wrap-multilines
```
pyphilia, line 184: You just need to add `()` around the jsx and remove this eslint comment.

src/components/parameter/FieldSelector.js
```diff
@@ -110,8 +98,69 @@ const FieldSelectorTree = ({ name, field, onChange, disabled }) => {
...
+    const { properties } = field;
+
+    const newChildren = Object.entries(properties).map(([key, childField]) => (
+      <FieldSelectorTree
+        key={key}
+        name={key}
+        field={childField}
+        disabled={disabled}
+        onChange={(updatedField) => {
+          // eslint-disable-next-line no-unused-expressions
```
pyphilia, line 114: Remove?

src/components/validation/AddValidation.js
```diff
@@ -0,0 +1,440 @@
...
+      history: { goBack },
+      dispatchCreateValidation,
+    } = this.props;
+    const { sourceId, selectedValidationAlgorithms, schemaId } = this.state;
+    dispatchCreateValidation({
+      sourceId,
+      algorithms: selectedValidationAlgorithms,
+      schemaId,
+    });
+    goBack();
```
pyphilia, line 169: Maybe use `push` to the validation page instead.

src/components/validation/AddValidation.js
```diff
@@ -0,0 +1,440 @@
...
+          id={buildAddValidationDatasetOptionId(id)}
+          value={id}
+          key={id}
+          className={classes.menuItem}
+        >
+          <Grid container alignItems="center" spacing={1}>
+            <Grid item>{name}</Grid>
+            {schemaIds?.map((schemaId) => (
+              <Grid item key={schemaId}>
+                <SchemaTag schema={schemas.get(schemaId)} />
```
pyphilia, line 213: Would it make sense for `SchemaTag` to process the `schemaId` itself? It would fetch `schemas` from the database and have `schemaId` as prop.

src/components/validation/AddValidation.js
```diff
@@ -0,0 +1,440 @@
...
+    const { algorithmId } = this.state;
+
+    const algorithmLabel = t('Validation Algorithm');
+
+    return (
+      <FormControl variant="outlined" className={classes.formControl}>
+        <InputLabel>{algorithmLabel}</InputLabel>
+        <Select
+          id={ADD_VALIDATION_ALGORITHMS_SELECT_ID}
+          labelId="algorithm-select"
```
pyphilia, line 269: Remove? Or add the id to the corresponding label.

src/reducers/executionsReducer.js
```diff
@@ -77,6 +79,14 @@ export default (state = INITIAL_STATE, { type, payload }) => {
           executions.findIndex(({ id: exId }) => exId === payload),
         ),
       );
+    case CREATE_VALIDATION_SUCCESS:
+      return state.update('executions', (executions) =>
+        payload.executions.reduce(
+          (executionsList, execution) =>
+            addExecutionToList(execution)(executionsList),
+          executions,
+        ),
```
pyphilia, line 88: Would it possible to use `forEach` here? I feel that using `reduce` is a bit complicated, but maybe I'm missing something.

pyphilia (changes requested):
Thanks for your work! Some general comments:
- You can change the algorithm's type but it doesn't change the template. Is it intended?
- Maybe let's delete the `Size` column in Algorithms. It will save a bit a space in this table.
- It doesn't make  sense to add twice the same validation algorithms when validating, right? If it is easy to prevent such behaviour please add it. Otherwise let's have an issue for that.
- Did you test a long running validation algorithm? I had some issues with the remaining running status when quitting the application, maybe add a test for this. But since your implementation is based on execution, you shouldn't have any problem.

src/components/execution/ExecutionTable.js
pyphilia: I prefer putting `type` in `algorithm`.

public/shared/utils.js
```diff
@@ -9,7 +9,7 @@
  * @return {object} - schema with selected and expanded values for each property
  */
 const setFieldSelectorAttributes = (schema, selected, expandUntilDepth) => {
-  if (!schema || !schema.properties) {
+  if (!schema) {
```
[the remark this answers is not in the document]
Painguin: I guess, if the root-level `type` is `object`, then there would be an error.  
But if the rool-level `type` is `array`, then `properties` is actually not required.

I've added internal checks to saveguard against bad schema definitions

src/components/common/Table.js
Painguin: Yes, I wanted to allow a cell to span two columns with `colSpan` but then we need another column to occupy 0 columns which I do by filtering out.

This is used in `Validation.js` where the Validation result and status are defined with a single cell spanning over both of them.

src/components/parameter/FieldSelector.js
Painguin: Same as before, I included internal checks to deal with that possibility when the schema refers to an object

src/components/parameter/FieldSelector.js
Painguin: I have an issue where `yarn prettier:write` removes those `()`... I didn't find a workaround...

src/reducers/executionsReducer.js
Painguin: A `forEach` would mean pushing sequentially each execution into the same executions list, which doesn't work in the immutable context (it returns an updated copy instead).

But you are right that `reduce` is too complicated there, I found `list.concat` is a lot more appropriate

src/components/validation/AddValidation.js
Painguin: I think it would make sense. The only issue is the `...` schema tag that shows up when there are more than 3 schemas. It is defined on the fly and is not in `schemas`. 

possible workarounds:
- have a different component for the `...` schema tag
- deal with it through the props somehow. Maybe check if id is some predefined constant

What do you think?

src/components/parameter/FieldSelector.js
pyphilia: I guess you can just factor out the component then?

src/components/validation/AddValidation.js
pyphilia: Isn't it possible to check depending on the number of tags? We can also create a whole component `SchemaTags` which receive the schemaIds and deal with the `...`? We can discuss live about it.

pyphilia: Isn't it possible to check depending on the number of tags? We can also create a whole component SchemaTags which receive the schemaIds and deal with the ...? We can discuss live about it.

public/app/listeners/createExecution.js
```diff
@@ -13,6 +13,30 @@ const {
 } = require('../../shared/types');
 const logger = require('../logger');
 
+const createExecutionObject = (
```
pyphilia, line 16: I guess we can call it `createExecutionInDb`

src/components/execution/ExecutionLog.js
```diff
@@ -30,18 +28,8 @@ const ExecutionLog = ({ id }) => {
 
   // fetch execution to update log every set interval
```
pyphilia, line 29: Remove this comment.

src/components/validation/Validation.js
```diff
@@ -0,0 +1,482 @@
...
+    } = this.props;
+    push(ADD_VALIDATION_PATH);
+  };
+
+  handleDelete = ({ id }) => {
+    const { dispatchDeleteValidation, t } = this.props;
+    dispatchDeleteValidation({ id, name: t('this validation') });
+  };
+
+  handleResultView = (resultId) => {
```
pyphilia, line 139: You don't seem to use it.

src/components/validation/Validation.js
```diff
@@ -0,0 +1,482 @@
...
+    const {
+      history: { push },
+    } = this.props;
+    push(buildResultPath(resultId));
+  };
+
+  handleSourceView = (sourceId) => {
```
pyphilia, line 146: Would it be possible to use `DatasetViewButton`?

pyphilia (approved):
Nice, almost there! Merge once you checked my comments.

hagoptaminian (commented):
Very interesting feature and nicely done. A few suggestions other than the one discussed:

- In general, considering using _Validate_ rather than _Execute_ in places related to Validation. Specifically: (1) On the 'Validate dataset' (add validation) page, the button, (2) In the Validation table: 'An error occurred during **execution**', 'View **execution**', (3) In the validation log page (title is Execution).
- 'Validate **D**ataset' as the title of the page (capital D)
- In the table, clicking on a link under the 'Validation result' header takes one to the validation algorithm - not sure this makes sense. Either rename header to 'Executed Validation' or 'Validation Performed', or I would expect 'Validation Result' to take me to the validation log. (also capitalize both letters in table header).
- Should we go with Validation**s** in the side menu and as the title of the page? To be consistent with other headings.
- Give the + button in 'Validate dataset' page a tooltip: 'Add algorithm' or 'Add algorithm to queue'. Basically, it took me a while to realize that I can select multiple validations at once. I thought the + button just led to an expanded view of the algorithm to select parameters. Then in the expanded view instead of 'Remove' you can use 'Remove algorithm'.
- On the 'Validate dataset' page, consider adding an 'information icon' tooltip at the top right (in line with the heading) which when hovered over explains what validations are. Something like: 'Use this feature to test whether datasets have been anonymized as expected.'
- On the 'Validate dataset' page, I find the + button a bit awkwardly placed and the UI a bit off... but perhaps not worth worrying about now.

For the record I feel there is room to simplify the Validation workflow to be more beginner friendly, but maybe that's for another time.

hagoptaminian (approved):

PR #165, opened 2021-04-28
feat: support csv files and add dataset csv view

Painguin:
Support csv and xlsx file formats. A Dataset with a csv and xlsx format is converted into an array of json objects for its use in the app.

In `DatasetScreen.js`, one can select between a json view and a csv (tabular) view. The csv view is only shown if the dataset supports it (array of json objects).

When exporting a dataset, one can select between the json a csv (if dataset supports it) format.

The generic algorithms (`hash fields`, `suppress fields`, `shuffle fields`) should all work with imported csv's (after generating a schema from them)


src/components/dataset/DatasetScreen.js
```diff
@@ -125,42 +219,54 @@ class DatasetScreen extends Component {
...
-                id={DATASET_NAME_ID}
-                className={classes.title}
-                variant="h4"
-                align="center"
-              >
-                {datasetName}
-              </Typography>
-            </Grid>
-            <Grid item xs={6}>
+            <Grid item xs={showTabular && !isGraasp ? 10 : 6}>
```
pyphilia, line 241: Can it be tabular and graasp?

src/components/dataset/DatasetScreen.js
```diff
@@ -125,42 +219,54 @@ class DatasetScreen extends Component {
...
-                  id={datasetId}
-                  content={datasetContent}
-                  collapsed={2}
-                  editEnabled
-                />
+              <Paper
+                className={format === FILE_FORMATS.JSON && classes.content}
+              >
+                {showTabular ? (
+                  this.renderTabularView()
```
pyphilia, line 257: Would it be clearer to have a separated component for the table view?

src/components/dataset/DatasetScreen.js
```diff
@@ -87,11 +103,87 @@ class DatasetScreen extends Component {
...
+    );
+  };
+
+  renderTabularView = () => {
+    const { datasetContent, t, isTabular } = this.props;
+
+    if (!isTabular) {
+      return (
+        <Alert severity="error">
+          {t("The dataset can't be displayed as a csv.")}
```
pyphilia, line 154: Maybe I would just disable to table view button if the dataset cannot be displayed as a table.

test/dataset.test.js
```diff
@@ -202,4 +203,27 @@ describe('Datasets Scenarios', function () {
       );
     }),
   );
+
+  it(
+    'Add, delete CSV dataset',
```
pyphilia, line 208: Add a test for xlsx as well?

pyphilia: Should you add tests for the view as well?

pyphilia (changes requested):
It works really well! I left some comments, feel free to pin me for questions!

public/app/listeners/loadDataset.js
```diff
@@ -1,6 +1,7 @@
 const path = require('path');
 const fs = require('fs');
 const ObjectId = require('bson-objectid');
+const XLSX = require('xlsx');
```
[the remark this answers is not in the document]
Painguin: Most of the examples I've seen online use either `XLSX` or `xlsx`. Should I still go with `Xlsx`?

public/app/listeners/loadDataset.js
```diff
@@ -12,25 +13,53 @@ const {
...
+  }
 
   let schemaIds;
+  let isTabular;
   try {
     const content = fs.readFileSync(destPath, 'utf8');
     const jsonContent = JSON.parse(content);
     schemaIds = detectSchemas(jsonContent, db) || [];
+
+    isTabular = validateSchema(jsonContent, ARRAY_OF_JSON_SCHEMA);
```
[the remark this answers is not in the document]
Painguin: Two corrections: we have to look if the data root type is an array and if each of the elements is an object. When modifying a dataset through the dataset screen and when creating a new result (after executing an algorithm), the isTabular value is recomputed again. So the only issue would be if the user goes himself in the directory to modify the dataset or if it gets corrupted somehow.

The alternative you propose is what I was going for first. But when we export a dataset, I need to check if the dataset is tabular to propose the `Export as CSV` option and that implies fetching the dataset content from the backend. This looked cumbersome and went with the `isTabular` solution

pyphilia (approved):
Nice! Just two minors comments, and you can merge!

